PR #37, opened 2021-09-15
Feat jest

Sebastien-Ahkrin:


package.json
```diff
@@ -31,7 +32,8 @@
   },
   "dependencies": {
     "@emotion/react": "^11.4.1",
-    "immer": "^9.0.6"
+    "immer": "^9.0.6",
+    "@testing-library/jest-dom": "^5.14.1"
```
targos, line 36: dev dependencies

package.json
```diff
@@ -41,19 +43,33 @@
...
     "@zakodium/eslint-config": "^3.0.3",
     "babel-loader": "^8.2.2",
     "eslint": "^7.32.0",
+    "jest": "^27.2.0",
     "prettier": "^2.3.2",
     "react": "^17.0.0",
     "react-dom": "^17.0.0",
     "rimraf": "^3.0.2",
+    "ts-jest": "^27.0.5",
+    "ts-node": "^10.2.1",
```
targos, line 62: seems unused?

package.json
```diff
@@ -21,6 +21,7 @@
     "eslint-fix": "npm run eslint -- --fix",
     "prettier": "prettier --check .",
     "prettier-write": "prettier --write .",
+    "test-only": "jest --watch",
```
targos, line 24: Only `jest` in `test-only`.
We can have another script with [`--watchAll`](https://jestjs.io/docs/cli#--watchall)

PR #65, opened 2021-10-04
fix: add zIndex on modals

Sebastien-Ahkrin:


src/layout/Modal.tsx
```diff
@@ -23,6 +23,7 @@ export function Modal(props: ModalProps) {
           position: 'absolute',
           top: 0,
           left: 0,
+          zIndex: 90,
```
targos, line 26: I think we should use a portal instead (add the modal at the end of the DOM so it's in front of everything)

src/layout/Modal.tsx
```diff
@@ -4,7 +4,7 @@ import { ReactElement, ReactNode } from 'react';
 import { FaTimes } from 'react-icons/fa';
 
 interface ModalProps {
-  children: ReactElement;
+  children: Array<ReactElement>;
```
targos, line 7: this came back?

Sebastien-Ahkrin: yes, used for Modal.XXXX

PR #71, opened 2021-10-06
Toolbar overflow

hamed-musallam:
show toolbar in a better way when not enough room for items, in this way i only show the items that fit the available space, and what remains i show it once the user click on the More button (3 dots icon) at the bottom of the toolbar


lpatiny:
@hamed-musallam thanks for this PR. We are exploring the possibility to simply deal with the overflow using CSS in order to keep the code simple and keep this PR in stand-by for the moment.

targos:
Fixed in https://github.com/cheminfo/analysis-ui-components/commit/6c038ae97dafa868c7b00a63c9640e93d00f071f

https://cheminfo.github.io/analysis-ui-components/?path=/story/layout-toolbar--vertical-toolbar

PR #80, opened 2021-11-10
Fix SplitPane movement

wadjih-bencheikh18:


lpatiny:
It seems to me that this does not fix the problem (at least not in the preview).

What we expect as behaviour is something like 



You click on the splitter and you move the mouse. The mouse can leave the splitter but when the splitter moves the mouse is expected to be on it. In the change you did if the mouse leaves the parent div the control is lost and this is not what we expect neither. Once we 'mousedown' on the splitter we keep the control on it even if we quit the parent div for as long as we keep the mousedown (only exception is if we leave the browser window).

wadjih-bencheikh18:
It wasn't clear in the issues
Aaaaaa okey now i understand the problem thank you

package.json
```diff
@@ -40,16 +40,21 @@
...
     "@storybook/addon-actions": "^6.3.12",
     "@storybook/addon-essentials": "^6.3.12",
     "@storybook/addon-links": "^6.3.12",
     "@storybook/addon-storysource": "^6.3.12",
     "@storybook/react": "^6.3.12",
+    "@testing-library/dom": "^8.11.1",
+    "@testing-library/jest-dom": "^5.15.0",
+    "@testing-library/react": "^12.1.2",
+    "@testing-library/user-event": "^13.5.0",
+    "@types/jest": "^27.0.2",
```
targos, line 54: Please remove these dependencies. They were here in an older version of the project but are not used anymore.
After this is done, make sure you have npm 7 or 8, then delete the package-lock.json, run `npm install`, and commit the new package-lock.json.

wadjih-bencheikh18: i deleted those lines and run npm install => i got  `55 vulnerabilities (3 low, 15 moderate, 35 high, 2 critical)`
my npm version is 8.1.0

wadjih-bencheikh18: and i updated all packages before install (so this is why package-lock) changed

wadjih-bencheikh18:
@lpatiny the issus is fixed now

lpatiny:
It is much better indeed. There are still some small shifts and I wonder if it is easy to fix. Otherwise we can keep it like that 


.

stropitek:
I can confirm that there are shifts. If you start jiggling your mouse fast around where the the split ends, you'll have large shifts.

wadjih-bencheikh18:
This shifts is due to math.floor i think 
I can try fix it

wadjih-bencheikh18:
@lpatiny it's better now

wadjih-bencheikh18:
@lpatiny it works perfectly now
can you check please?

wadjih-bencheikh18:
If the orientation is horizontal and the cursor go out the parent element vertically
is there any details in this case?

PR #82, opened 2021-11-14
Add drag & drop zone

wadjih-bencheikh18:


wadjih-bencheikh18:
done

lpatiny:
In fact we have 2 different cases. The first one is the one you did. Nothing is currently loaded and we can drag and drop a new spectrum


es

There is also another case in which there is already a spectrum and we still one to add one more. In fact the drop component is a wrapper and by default it would be invisible. However if you drag a file over it you should see that you can drop a file in the zone.



I would add a property 'visible' to distinguish between the 2 behaviour.

wadjih-bencheikh18:
okey,Thnx for details

wadjih-bencheikh18:
states added,
i used to test a place holder image (u can see it as a comment if u want to visualise the active state)

src/layout/DropZone.tsx
```diff
@@ -0,0 +1,144 @@
...
+                  background-color: red;
+                  color: white;
+                `}
+              >
+                {' '}
+                Test{' '}
+              </div>
+            </div>
+          ) : (
+            ''
```
stropitek, line 80: Please use `null` when you want to render empty elements

stropitek:
@wadjih-bencheikh18 are you finished?

I think the part about the dropzone being just a wrapper that renders it's children isn't implemented yet

targos:
> I think the part about the dropzone being just a wrapper that renders it's children isn't implemented yet

To be clear, this should be an option.

src/layout/DropZone.tsx
```diff
@@ -0,0 +1,144 @@
...
+  return (
+    <>
+      <div
+        {...getRootProps()}
+        css={css`
+          overflow: hidden;
+          position: relative;
+          width: ${props ? props.width : '100%'};
+          min-width: 400px;
+          min-height: 300px;
```
stropitek, line 39: Avoid adding too many default styles to the component. It's better to let the user of the component add their own styles.

wadjih-bencheikh18:
> 

yes i finished, thank you

src/layout/DropZone.tsx
wadjih-bencheikh18: deleted

src/layout/DropZone.tsx
wadjih-bencheikh18: fixed

wadjih-bencheikh18:
> > I think the part about the dropzone being just a wrapper that renders it's children isn't implemented yet
> 
> To be clear, this should be an option.

yes there is two states as luc mentioned

stropitek:
If you drag a file over the dropzone in the story, the rendered overlay doesn't look very nice.

> yes there is two states as luc mentioned

Can you explain? I don't see it in the code or in the story. We would need another story where:
- Some content is rendered (let's say it's a blue div with a fixed height and width), wrapped in your component.
- The presence of the dropzone is not apparent (no text and dotted borders) before you start dragging on it.
- When you drag a file over the blue div, there is some overlay that shows that you can drop a file there.

wadjih-bencheikh18:
> If you drag a file over the dropzone in the story, the rendered overlay doesn't look very nice.
> 
> > yes there is two states as luc mentioned
> 
> Can you explain? I don't see it in the code or in the story. We would need another story where:
> 
> * Some content is rendered (let's say it's a blue div with a fixed height and width), wrapped in your component.
> * The presence of the dropzone is not apparent (no text and dotted borders) before you start dragging on it.
> * When you drag a file over the blue div, there is some overlay that shows that you can drop a file there.

stories added 
- i deleted border when active (we already have file blue div) 
- active you can drag an item and see the diffrence 
- empty is for the case when we don't have any file (you can also drag and item and see that we are in the active state)

wadjih-bencheikh18:
@targos could you please check if everything is great?

stropitek:
@wadjih-bencheikh18 for me we cannot merge this yet:

- The standalone version does not handle the click event
- The wrapper version has styling props that should not be there (e.g. `width`). It should not add / remove anything to the rendering of its children. If you'd render the children with or without the wrapper the result should be the same.

As I suggested when we discussed yesterday, you can create 2 separate components with a different api if it's needed.

src/layout/DropZone.tsx
```diff
@@ -0,0 +1,130 @@
...
+import React, { useCallback, useState } from 'react';
+import { useDropzone } from 'react-dropzone';
+import { MdCloudUpload } from 'react-icons/md';
+
+export default function DropZone(props: {
+  color?: string;
+  children?: JSX.Element;
+}) {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [alert, setAlert] = useState('');
```
targos, line 12: no-unused-vars errors should never be disabled. either use the variable, or remove it.

stories/DropZone.stories.tsx
```diff
@@ -7,25 +7,20 @@ export default {
   title: 'Layout/DropZone',
   args: {
     color: 'black',
-    width: '50%',
+    children: <div style={{ backgroundColor: 'blue' }}>hello world</div>,
```
stropitek, line 10: Please move the children to the story itself

src/layout/DropZone.tsx
```diff
@@ -0,0 +1,130 @@
...
+    <>
+      <div
+        {...getRootProps({ onClick: (event) => event.stopPropagation() })}
+        css={css`
+          overflow: hidden;
+          position: relative;
+          height: 200px;
+          width: 100%;
+          ${!active || isDragActive
+            ? `border: 2px dashed ${props.color ? props.color : 'black'};`
```
targos, line 43: please create a variable above like this to avoid the repetition:

```js
const { color = 'black' } = props;
```

src/layout/DropZone.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const { getRootProps, getInputProps, isDragActive } = useDropzone({ onDrop });
+
+  return (
+    <>
+      <div
+        {...getRootProps({ onClick: (event) => event.stopPropagation() })}
+        css={css`
+          overflow: hidden;
+          position: relative;
+          height: 200px;
```
targos, line 40: the height should be 100% too (the component should adapt to where it is)

src/layout/DropZone.tsx
```diff
@@ -0,0 +1,130 @@
...
+            font-weight: 600;
+            width: 100%;
+            padding: 10px;
+          `}
+        >
+          {active ? (
+            <div
+              css={css`
+                width: 100%;
+                z-index: -1;
```
targos, line 65: please don't use z-index to implement this

src/layout/DropZone.tsx
```diff
@@ -0,0 +1,130 @@
+/** @jsxImportSource @emotion/react */
+import { css } from '@emotion/react';
+import React, { useCallback, useState } from 'react';
+import { useDropzone } from 'react-dropzone';
+import { MdCloudUpload } from 'react-icons/md';
```
targos, line 5: please use an icon from font-awesome (react-icons/fa)

src/layout/DropZone.tsx
```diff
@@ -0,0 +1,130 @@
...
+      reader.onerror = () => setAlert('file reading has failed');
+      reader.onload = () => {
+        setActive(true);
+        // Do whatever you want with the file contents
+        //const binaryStr = reader.result;
+        setAlert('file uploaded');
+        //console.log(binaryStr);
+      };
+      reader.readAsArrayBuffer(file);
+    });
```
targos, line 29: we should not read the files here. the component should accept a callback prop `onDrop: (files: TypeOfAcceptedFiles) => void`

src/layout/DropZone.tsx
wadjih-bencheikh18: fixed

src/layout/DropZone.tsx
wadjih-bencheikh18: done

src/layout/DropZone.tsx
wadjih-bencheikh18: okey deleted

src/layout/DropZone.tsx
wadjih-bencheikh18: done

stories/DropZone.stories.tsx
wadjih-bencheikh18: no need to the children now

src/layout/DropZone.tsx
wadjih-bencheikh18: added

src/layout/DropZone.tsx
wadjih-bencheikh18: done

stories/DropZone.stories.tsx
stropitek: I'd like to see what is rendered inside as children in the story

image

src/layout/DropZone.tsx
```diff
@@ -0,0 +1,125 @@
+/** @jsxImportSource @emotion/react */
+import { css } from '@emotion/react';
+import React, { useCallback, useState } from 'react';
+import { useDropzone } from 'react-dropzone';
+import { FaCloudUploadAlt } from 'react-icons/fa';
+
+export default function DropZone(props: {
+  color?: string;
+  children?: JSX.Element;
+  Drop: (files?: unknown) => void;
```
stropitek, line 10:
```suggestion
  onDrop: (files: File[]) => void;
```

stories/DropZone.stories.tsx
wadjih-bencheikh18: children arg added in the active story
but the display of Jsx element is not easily readable , is there any way to improve the children arg display in this story?

wadjih-bencheikh18:
@stropitek is there any other changes needed?

wadjih-bencheikh18:
@stropitek is there any other changes needed?

stropitek:
> We could define the list of allowed extension, a message when you are able to drop the file (the right extension) and a message when you are not allowed to drop the file (wrong extension).

I don't think this has been done or tested. But you can do this in another PR. I'm merging this one already.

PR #103, opened 2022-03-09
feat: Render components from layout object

redhaam:
I need to know the list of all the components which the user can pick from when declaring the layout. Also, I didn't get what was the component `BlueBackground` in the example of the layout provided.
I couldn't find a way to have a dynamic types for the properties.


lpatiny:
We will have the default components that are managed by the map you did but we should plan also some external components that are given to the layout manager.

Please add property that allows this and create the following dummy components:
- BlueBackground (a div with a blue background)
- RedBackground (a div width a red background)
- ...

lpatiny:
How did you do to test that it works ? There is no storybook ?

Please add in the storybook a new category 'layout' that display the result of the example.

lpatiny:
Please also check the CI tests. 2 are failing.

PR #115, opened 2022-03-17
feat: add table

wadjih-bencheikh18:


lpatiny:
Thanks for your work and sorry for the late answer.

This idea is to have a folder 'renderer' that will contain some basic components that will be reused in all our projects. So the goal here is not to create a 'TD' component but to have the content of the TD.

In this folder we will have a file `index.ts` that reexports all those renderer basic components.

If possible the style of the component will directly be placed in the component it self.

Could you move in this directly and update the table based on those components ?

wadjih-bencheikh18:
> Could you move in this directly and update the table based on those components ?
>


Yeah no problem

src/layout/Table/Cells.tsx
```diff
@@ -0,0 +1,47 @@
...
+export function NumberCell({
+  value,
+  fixed,
+}: {
+  value?: number;
+  fixed?: number;
+}) {
+  return <div>{value ? (fixed ? value.toFixed(fixed) : value) : ''}</div>;
+}
+export function TextCell({ value }: { value?: string }) {
```
lpatiny, line 42: Cell should be nowhere in the component name. Those are generic components.

wadjih-bencheikh18: fixed

lpatiny:
@wadjih-bencheikh18 could you check this PR so that the tests pass ?

wadjih-bencheikh18:
> @wadjih-bencheikh18 could you check this PR so that the tests pass ?

Fixed

PR #121, opened 2022-04-27
feat: specify whether the side panel is closed or not

hamed-musallam:


src/layout/SplitPane.tsx
```diff
@@ -21,6 +21,7 @@ export interface SplitPaneProps {
   orientation?: SplitOrientation;
   sideSeparation?: SideSeparation;
   initialSeparation?: InitialSeparation;
+  isSidePaneClosed?: boolean;
```
targos, line 24: initialClosed

hamed-musallam: @targos can you publish a new release?

PR #122, opened 2022-04-28
feat: Basic dynamic layout

hamed-musallam:


hamed-musallam:
@targos 
The Accordion component when you double click on it, closes all the Accordion components, it looks like they share the same state, Should we use the context at the level of each Accordion component, or we Should keep it globally and in this case, we need to differentiate between the accordion by identifier to be sure we only close the accordion items that related to it, what do you think is the best way?

targos:
Basic LayoutManager was merged in https://github.com/zakodium-oss/analysis-ui-components/commit/c82980ce7e6a6f16ff00dda9734b730a8e98cd16.
We will discuss the next steps next week.

hamed-musallam:
@targos 
Sorry I did not get it, i can not see the changes that I did in the merge that you did https://github.com/zakodium-oss/analysis-ui-components/commit/c82980ce7e6a6f16ff00dda9734b730a8e98cd16. do you mean no need for the code that i did in ```hamed-musallam:dynamic-layout```

targos:
The changes in your branch are in direct conflict with the PR that I merged. If there are missing things they should be done on top of the main branch now (in a new PR)

PR #141, opened 2022-05-19
feat!: create MeasurementPlot component

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/140


lpatiny:
When changing the size in the storybook it crash.

Please use as example this new file and delete the previous json:

[ir_0.json.zip](https://github.com/zakodium-oss/analysis-ui-components/files/8749034/ir_0.json.zip)

src/components/MeasurementPlot.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <Plot width={width} height={height}>
+      <Heading title={title} />
+      {data.map(
+        ({ variables: { [xVariableName]: x, [yVariableName]: y } }, i) => [
+          <LineSeries
+            // eslint-disable-next-line react/no-array-index-key
+            key={i}
+            data={xyToXYObject({
+              x: x.data.splice(dataIndex),
```
lpatiny, line 48: This is not what we would like here. The dataIndex is expected to select the data at the top level (sorry I used 'data' twice in the object hierarchy).

```
  "title": "Mattson Instruments",
  "data": [
    {
````

In the new example data is an array of 2 elements and it should therefore work with dataIndex 0 and 1.

On a side node please take care about the difference between `splice` and `slice`. Splice is 'in-place' meaning you actually change the original array and therefore there is a side effect we nearly never want. `slice` will make a copy (will take memory and could be slow).

Also if we need to send part of an array to a function and the array is a typed array we can use `subarray`. This will not be a copy but just a new pointer to the part of the typed array. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/subarray

**Comment by lpatiny** [comment]
The new dataset also contains more variables and int he storybook we should be able to select among the 4 available.

**Review by lpatiny — CHANGES_REQUESTED** [review]

File: src/components/MeasurementPlot.tsx [hunk 2]
```diff
@@ -0,0 +1,66 @@
...
+export function MeasurementPlot(props: MeasurementPlotProps) {
+  const {
+    measurment,
+    dataIndex = 0,
+    xVariableName = 'x',
+    yVariableName = 'y',
+    width = 800,
+    height = 400,
+  } = props;
+  const { title = '', data } = measurment;
```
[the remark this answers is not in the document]
wadjih-bencheikh18: fixed

src/components/MeasurementPlot.tsx
wadjih-bencheikh18: fixed,
thank you

lpatiny (approved):
The functionalities are ok.

lpatiny:
@targos Could you check the implementation ? Functionality is ok.

After merge I think @wadjih-bencheikh18 could improve the component:
- https://github.com/zakodium-oss/analysis-ui-components/issues/144

And wrap it in another component to be able to select what to display:
- https://github.com/zakodium-oss/analysis-ui-components/issues/145

lpatiny:
@wadjih-bencheikh18  Could you still take care that if there is a wrong variable name there is a message explaining this variable is unknown rather than a crash ?

In the storybook just change the variable to reproduce this problem.

TypeError: Cannot read properties of undefined (reading 'data')
    at MeasurementPlot (https://deploy-preview-141--analysis-ui-components.netlify.app/storybook/assets/iframe.2d27371f.js:1154:132130)
    at Ch (https://deploy-preview-141--analysis-ui-components.netlify.app/storybook/assets/iframe.2d27371f.js:178:57361)

PR #147, opened 2022-05-23
feat: close the split panel based on the minimum size property 

hamed-musallam:
closed the panel on load when the panel size is less than or equal to the minimum size,
This action is not effective anymore once the user touched the splitter by clicking on it.

### DIFF
--- a/src/components/SplitPane.tsx
+++ b/src/components/SplitPane.tsx
@@ -7,6 +7,9 @@ import {
   useContext,
   useRef,
   useState,
+  useCallback,
+  useLayoutEffect,
+  forwardRef,
 } from 'react';
 
 import { useSplitPaneSize } from './hooks/useSplitPaneSize';
@@ -41,6 +44,11 @@ export interface SplitPaneProps {
    * @default false
    */
   initialClosed?: boolean;
+  /**
+   * Define when the split panel should be closed by setting the minimum size in Pixel 'px'
+   * @default undefined
+   */
+  minimalSize?: number;
   /**
    * @default () => null
    */
@@ -108,12 +116,15 @@ export function SplitPane(props: SplitPaneProps) {
     sideSeparation = 'start',
     initialSeparation = '50%',
     initialClosed = false,
+    minimalSize,
     onChange = () => null,
     children,
   } = props;
 
   const [isSidePaneClosed, toggle] = useToggle(initialClosed);
   const parentRef = useRef<HTMLDivElement>(null);
+  const panelRef = useRef<HTMLDivElement>(null);
+  const touchedRef = useRef<boolean>(false);
 
   const [[size, type], setSize] = useState(() => {
     const [, value, type] = /(?<value>^\d+)(?<type>.+)$/.exec(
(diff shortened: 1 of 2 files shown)


targos:
Why did you not use a resize observer as suggested in the issue? AFAIK this implementation will only check the size on the first render and not react to window size changes.

src/components/SplitPane.tsx
```diff
@@ -207,3 +249,47 @@ function getSize(
...
+  );
+}
+
+interface SplitPaneInnerProps {
+  style: CSSProperties;
+  value: boolean;
+  children: React.ReactNode;
+}
+
+const SplitPaneInner = forwardRef<HTMLDivElement, SplitPaneInnerProps>(
```
targos, line 284: I think we should avoid to overuse `forwardRef`, especially in internal components. You can pass the ref with a simple prop, for example `innerRef`, or `divRef`.

hamed-musallam:
> Why did you not use a resize observer as suggested in the issue? AFAIK this implementation will only check the size on the first render and not react to window size changes.

@targos 
I used it at the beginning but when I thought about where we are going to use this component, for example, in NMRium, the component will re-render after we change the window size so I thought no need to use the resize observer

src/components/SplitPane.tsx
hamed-musallam: Yes, I agree, you need it as internal props to be sure that it is updated. I did it like that because the component will re-render, for example, if we use the component in another component like NMRium.

But you are right, it is a general component, and will use it not only in NMRium , I will make the changes.

targos:
>  the component will re-render after we change the window size

There's no way to be sure about that. Even if it's true today, it can change in the future for any reason and break the assumption.

src/components/SplitPane.tsx
```diff
@@ -108,12 +115,14 @@ export function SplitPane(props: SplitPaneProps) {
...
     initialSeparation = '50%',
     initialClosed = false,
+    minimalSize,
     onChange = () => null,
     children,
   } = props;
 
   const [isSidePaneClosed, toggle] = useToggle(initialClosed);
   const parentRef = useRef<HTMLDivElement>(null);
+  const touchedRef = useRef<boolean>(false);
```
stropitek, line 125: You could get rid of this ref by allowing your `isSidePaneClosed` to take a third initial value (`null` for example)

src/components/SplitPane.tsx
```diff
@@ -133,6 +142,59 @@ export function SplitPane(props: SplitPaneProps) {
...
+        const { width: parentWidth, height: parentHeight } =
+          parentRef.current.getBoundingClientRect();
+
+        const parentDimension = [parentWidth, parentHeight];
+        const panelDimension = [width, height];
+        const sizeIndex = orientation === 'horizontal' ? 0 : 1;
+        const size =
+          sideSeparation === 'start'
+            ? panelDimension[sizeIndex]
+            : parentDimension[sizeIndex] - panelDimension[sizeIndex];
```
stropitek, line 170: There is probably a bug somewhere else if you absolutely need use the size of different divs depending on the `sideSeparation`

src/components/SplitPane.tsx
```diff
@@ -133,6 +142,59 @@ export function SplitPane(props: SplitPaneProps) {
...
+      return cssStyles.item(
+        isFinalClosed,
+        sideSeparation === side,
+        orientation,
+        size,
+        type,
+      );
+    },
+    [isFinalClosed, orientation, sideSeparation, size, type],
+  );
```
stropitek, line 156: There is no need to have a `useCallback` here. It created an additional layer of indirection which makes the code more difficult to read and understand

I think you can just call `cssStyles.item` directly and put the styles in a variable

stropitek (commented):
Hi @hamed-musallam!

It seems to me that there is an overuse of `useRef`

`useRef` should be used with parsimony because it goes against the core principles of declaritivity and immutability in react

src/components/SplitPane.tsx
targos: I think this ref is OK. The "isTouched" information is not needed outside the component and shouldn't trigger a rerender when it changes.

src/components/SplitPane.tsx
hamed-musallam: I assume I only track the change of the size for one side of the splitPane so if we change the sideSpeapartion === "end" then we need to call the size ``` parentDimension[sizeIndex] - panelDimension[sizeIndex]```.

but yes we can get rid of the parentRef  from the resizer handler function if we are going to pass it to the onResize call-back based on the sideSeparation

targos: @lpatiny What size should be tracked in your opinion? The total (content+separator+panel), the content, or the panel?

src/components/SplitPane.tsx
hamed-musallam: because it is called into two places based on a hardcoded sideSeparation value

stropitek: It's fine for me to keep it that way. But maybe we can remove the `useCallback` unless there is a performance issue linked to recreating this function on every render

src/components/SplitPane.tsx
targos: Why not pass the resize observer's `ref` conditionally to the correct child?

stropitek: Right that does not work but I think you can follow Michael's suggestion to pass the ref to the one or the other based on `sideSeparation`

stropitek:
The story "minimal size" story could be fluid (expand to fill the available height) also in the vertical direction so that the feature can be tested with a vertical separation

stropitek:
LGTM

@targos can we merge?

PR #150, opened 2022-06-02
feat: Measurement explorer

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/145


src/components/color-picker/helpers/checkboard.js
```diff
@@ -5,7 +5,9 @@ export function render(c1, c2, size, serverCanvas) {
     return null;
   }
   const canvas = serverCanvas
-    ? new serverCanvas()
+    ? // todo : fix the error "A constructor name should not start with a lowercase letter"
```
targos, line 8: the solution here is to rename the variable to `ServerCanvas`

wadjih-bencheikh18: fixed , thank you

lpatiny:
@wadjih-bencheikh18 Thanks it is what I wanted.

It is a nice idea to forbid to select twice the same variable but it may be not that practical in fact.

Let's say you have 2 variables and you display 'x' versus 'y' and you want to switch to 'y' versus 'x'. I have the impression it is not feasible currently.

Please just remove this extra feature and allow to select twice the same variable.

src/components/color-picker/helpers/checkboard.js
```diff
@@ -22,14 +22,14 @@ export function render(c1, c2, size, serverCanvas) {
...
-  const key = `${c1}-${c2}-${size}${serverCanvas ? '-server' : ''}`;
+export function get(c1, c2, size, ServerCanvas) {
+  const key = `${c1}-${c2}-${size}${ServerCanvas ? '-server' : ''}`;
 
   if (checkBoardCache[key]) {
     return checkBoardCache[key];
   }
 
-  const checkBoard = render(c1, c2, size, serverCanvas);
+  const checkBoard = render(c1, c2, size, ServerCanvas);
```
lpatiny, line 32: Why is this file not called 'checkBoard.js' but 'checkboard.js' ? I think it should use camelcase.

Please take care that when changing only the case it is better to use `git mv checkboard.js checkBoard.js` in order to avoid trouble with non case sensitive FS.

wadjih-bencheikh18: fixed

wadjih-bencheikh18:
> I have the impression it is not feasible currently.

I can add a button to reverse axis ( i think it will be better )

lpatiny:
In the header there is twice the xVariable

image

By default it is written x versus x and it should be y versus x. Moreover what is displayed is correct but does not corresponds to the selected variables.

lpatiny:
Adding button to reverse axis is a good idea. If it is quickly done you may do it but we should merge this branch ASAP.

wadjih-bencheikh18:
> In the header there is twice the xVariable

fixed thanks

wadjih-bencheikh18:
> If it is quickly done you may do it but we should merge this branch ASAP.

src/components/MeasurementExplorer.tsx
```diff
@@ -1,6 +1,7 @@
 /** @jsxImportSource @emotion/react */
 import { css } from '@emotion/react';
 import React, { useState } from 'react';
+import { CgArrowsExchangeAlt } from 'react-icons/cg';
```
targos, line 4: Can you please use either a unicode character, or import something from `react-icons/fa`. We don't want to use something than FontAwesome as icons.

PR #155, opened 2022-06-22
fix: enlarge the side panel does not appear anymore after shrinking it

hamed-musallam:


hamed-musallam:
@targos 
do you think having separate if statements are better than combining them into a single if statement like what we have here in this example? I mean from the code style view.



```ts
  useResizeObserver<HTMLDivElement>({
    onResize: ({ width, height }) => {
      const size = orientation === 'horizontal' ? width : height;
      if (
        size &&
        minimumSize &&
        !touchedRef.current &&
        ((size <= minimumSize && !isFinalClosed) ||
          (size > parenSizeRef.current && isFinalClosed))
      ) {
        const parentBounding = parentRef.current?.getBoundingClientRect();
        if (parentBounding) {
          parenSizeRef.current =
            orientation === 'horizontal'
              ? parentBounding.width
              : parentBounding.height;
        }
        toggle();
      }
    },
    ref: panelRef,
  });

```

targos:
> do you think having separate if statements are better than combining them into a single if statement like what we have here in this example? I mean from the code style view.

I think the code from the PR is easier to reason about but I would like to see some comments explaining the different cases because it's a bit complex.

hamed-musallam:
@targos 
Please, can you check it one more time

PR #162, opened 2022-06-23
Data

lpatiny:
Prepare the conversion of data in order to create a DataContext

The idea is that you can:
1.  getEmptyDataState()
2. append to the state a 'FileList'


src/components/context/data/DataState.ts
```diff
@@ -0,0 +1,48 @@
...
+  neta: Record<string, string>;
+  info: Record<string, string | number>;
+  data: {
+    variables: Record<string, MeasurementVariable>;
+  }[];
+}
+
+export type MeasurementKind = keyof Measurements;
+
+export const kindsLabel: Record<MeasurementKind, string> = {
```
lpatiny, line 41: @wadjih-bencheikh18 How should I do so that all the properties from 'MeasurementKind' are defined in this object and that there is an error if there is one extra property or missing property.

wadjih-bencheikh18: Using Record<MeasurementKind, string> is enough here

src/components/context/data/append.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Adds in the state the information that can be extracted from the fileList
+ * We can add 'processors' that based on the extension, filename, path or content
+ * could process the data before adding it in the state
+ */
+export async function append(
+  fileList: PartialFileList,
+  baseState: DataState,
+  options: any = {},
```
wadjih-bencheikh18, line 17: Is it possible to no use the type `any` here?

src/components/context/data/DataState.ts
```diff
@@ -0,0 +1,48 @@
...
+  other: {
+    entries: MeasurementBase[];
+  };
+}
+
+interface MeasurementBase {
+  id: string;
+  title?: string;
+  instrument?: Instrument;
+  neta: Record<string, string>;
```
wadjih-bencheikh18, line 32: I think this should be `beta`

src/components/context/data/processors/jcampProcessor.ts
```diff
@@ -0,0 +1,81 @@
...
+            title: measurement.title,
+            data: normalizeSpectra(measurement.spectra, kind),
+          });
+        }
+      }
+    }
+  }
+}
+
+function normalizeSpectra(spectra: any, kind: string) {
```
wadjih-bencheikh18, line 47: spectrum type in jcampconverter
```
/**
 * @typedef { Record<string, any> } Spectrum
 * @property {Record<string, number[]>} [data]
 * @property {number} [firstX]
 * @property {number} [lastX]
 * @property {number} [deltaX]
 * @property {number} [yFactor]
 * @property {number} [xFactor]
 * @property {number} [nbPoints]
 */
```

src/components/context/data/DataState.ts
lpatiny: I don't think that this solves the problem of a missing property.

src/components/context/data/DataState.ts
lpatiny: Actually it should be meta but indeed it was wrong

src/components/context/data/processors/jcampProcessor.ts
```diff
@@ -0,0 +1,81 @@
...
+import { v4 } from 'uuid';
+
+import { DataState, MeasurementKind } from '../DataState';
+import { irSpectrumEnhancer } from '../enhancers/irSpectrumEnhancer';
+
+const enhancers: Record<string, typeof irSpectrumEnhancer> = {
+  ir: irSpectrumEnhancer,
+};
+
+export async function jcampProcessor(
```
lpatiny, line 12: How can I specify that jcampProcessor is of type 'Processor' ?

targos: `export const jcampProcessor: Processor = async function (...) {}`

src/components/context/data/DataState.ts
lpatiny: @targos could you comment on this ?

targos: I'm not sure what you mean. If I forget a property there is an error:
image

targos:
I'm doing the merge with the main branch, but for next time please make sure you are up-to-date when you create your branch because here your base was from 1 month ago.

package.json
```diff
@@ -41,12 +43,17 @@
...
+    "filelist-utils": "^0.5.0",
     "immer": "^9.0.15",
+    "jcampconverter": "^9.0.2",
     "lodash": "^4.17.21",
     "react-dropzone": "^14.2.1",
     "react-shadow": "^19.0.3",
     "tinycolor2": "^1.4.2",
-    "use-resize-observer": "^9.0.2"
+    "use-resize-observer": "^9.0.2",
+    "uuid": "^8.3.2"
```
targos, line 56: Please remove this dependency (and `@types/uuid`) and use `@lukeed/uuid` that is much smaller, faster and already used in the project.

src/components/context/data/processors/jcampProcessor.ts
lpatiny: We will update this when jcampconverter will be migrate to typescript. Thanks !

PR #166, opened 2022-06-23
MeasurmentPlot new params

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/144
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/173

### DIFF
--- a/src/components/MeasurementExplorer.tsx
+++ b/src/components/MeasurementExplorer.tsx
@@ -65,7 +65,7 @@ export function MeasurementExplorer(props: MeasurementExplorerProps) {
               if (value !== undefined && !isNaN(value)) {
                 setInfo(({ flipHorizontalAxis }) => ({
                   flipHorizontalAxis,
-                  ...defaultInfo(0),
+                  ...defaultInfo(value),
                 }));
               }
             }}


lpatiny:
Here is the direct link: https://deploy-preview-166--analysis-ui-components.netlify.app/storybook/?path=/story/layout-measurementexplorer--control

Changing the variable (x / y) crash the system.

Could you also update the storybook to be able to activate / desactivate the different new options ?

wadjih-bencheikh18:
> Changing the variable (x / y) crash the system.

I added an error that you have to enter valid variable

lpatiny:
But do you know why the spectrum disappear ?

wadjih-bencheikh18:
> But do you know why the spectrum disappear ?
> 

Yes i'm now fixing it

lpatiny:
Playing around with zoom and change of variables crash

image

wadjih-bencheikh18:
> Playing around with zoom and change of variables crash

bug fixed
there's one problem that  we cannot use the same variable

lpatiny (changes requested):
There is a crash if you try to switch the axis

image

wadjih-bencheikh18:
> There is a crash if you try to switch the axis
>

this problem is due to a contrainte in react-plot that force the user to use x as horizontal and y to vertical

i fixed this problem here but changing axisId 

```
  xAxis = `${xVariableName}-x`
  yAxis =`${yVariableName}-y`
```


> there's one problem that we cannot use the same variable

also with this i could fix same variable problem

src/components/MeasurementPlot.tsx
```diff
@@ -72,6 +72,8 @@ function MeasurementComponent(props: MeasurementPlotProps) {
   } = props;
   const { title = '', data } = measurement;
 
+  const xAxis = useMemo(() => `${xVariableName}-x`, [xVariableName]);
+  const yAxis = useMemo(() => `${yVariableName}-y`, [yVariableName]);
```
targos, line 76: Strings are always unique, you don't need to memo them.

wadjih-bencheikh18: removed, thank you

PR #167, opened 2022-06-23
add flip btn in measurementExplorer

wadjih-bencheikh18:
close : https://github.com/zakodium-oss/analysis-ui-components/issues/159


lpatiny:
Here is the direct link to the preview: https://deploy-preview-167--analysis-ui-components.netlify.app/storybook/?path=/story/layout-measurementexplorer--control

But it does not work on my computer. I can not change the dataIndex anymore and clicking on flip axis does not work.

src/components/MeasurementExplorer.tsx
```diff
@@ -1,6 +1,8 @@
 /** @jsxImportSource @emotion/react */
 import { css } from '@emotion/react';
-import React, { useState } from 'react';
+import { faArrowsLeftRightToLine } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { useState } from 'react';
 import { FaExchangeAlt } from 'react-icons/fa';
```
targos, line 6: Why is there a mix of `react-icons/fa` and a new library?

targos: ok. I will open an issue to migrate react-icons/fa to this

lpatiny: For now we could rather use this icon: https://fontawesome.com/icons/arrows-left-right?s=solid

And remove { FontAwesomeIcon } from '@fortawesome/react-fontawesome';

Even if are only using react-icons/fa it seems easier to insert.

wadjih-bencheikh18: react-icons/fa is easy to insert but we are so limited in icons
there is much more icons in fontawesome

lpatiny: I created an issue to find out why it was not updated: https://github.com/react-icons/react-icons/issues/573. In the mean time we can use react-icons and hope that will update soon (or contribute to their project)

wadjih-bencheikh18:
@lpatiny  fortawesome dependencies removed

PR #169, opened 2022-06-25
Measurement panel

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/165


lpatiny:
Please:

1. Check why there is this semicolumn
image

2. Move this panel to the top (It should replace the 'Spectra' panel

3. Rebase your branch once https://github.com/zakodium-oss/analysis-ui-components/pull/171 is merged

src/components/context/data/DataState.ts
```diff
@@ -30,23 +30,31 @@ interface MeasurementBase {
...
-export const kindsLabel: Record<MeasurementKind, string> = {
+export const kindLabelMap: Record<MeasurementKind, string> = {
   ir: 'IR',
   raman: 'Raman',
   uv: 'UV',
   mass: 'Mass',
   nmr1h: 'NMR 1H',
   other: 'Other',
 };
+export const kindsLabel: MeasurementKind[] = [
```
lpatiny, line 50: Please don't create this new variable and when you need the properties just use `Object.keys(kindLabel)`. It should not be a limiting speed step. And even we we wanted to create this second array we should not write all the values but also use `Object.keys`.

lpatiny (changes requested):

src/components/context/data/DataState.ts
wadjih-bencheikh18: fixed ,
thank you

lpatiny (approved):
LGTM, I checked locally and the functionalities are there.

@targos could you merge this PR if the code is ok ? I could not understand why the preview on netify is not working.

targos:
The Netlify preview fails with a memory error during build. I think it's probably because of the big JSON files that are imported and parsed. Can you try to use the [`?raw` imports](https://vitejs.dev/guide/assets.html#importing-asset-as-string)? If it doesn't work, we will have to put them in the public folder and load with `fetch()`

wadjih-bencheikh18:
> If it doesn't work, we will have to put them in the public folder and load with `fetch()`

yes after tryin, it didn't work

PR #170, opened 2022-06-25
Improve dropzone story

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/163


stories/dropZone.stories.tsx
```diff
@@ -7,6 +7,8 @@ export default {
   title: 'Layout/DropZone',
   args: {
     color: 'black',
+    // eslint-disable-next-line no-alert
+    onDrop: () => alert('files uploaded'),
```
lpatiny, line 11: Is there a possibility to display the list of dropped files ?

Like in [this example](https://github.com/cheminfo/filelist-utils/blob/59e182c6649ebcf39df7880b47482190b18ac2b3/html/index.html).

Also could we display and edit the 'onDrop' callback in the storybook ?

lpatiny (changes requested):

lpatiny:
The goal here is really to quickly and easily be able to debug what is happening when you drop one file, a list of files or a folder. This is why we would need a callback that displays what was dropped, maybe in a JSON inspector if easy to do.

stories/dropZone.stories.tsx
wadjih-bencheikh18: > Is there a possibility to display the list of dropped files ?

yeah it's possible

lpatiny:
Please also rebase this branch.

stories/dropZone.stories.tsx
wadjih-bencheikh18: > Is there a possibility to display the list of dropped files ?

table created.



> Also could we display and edit the 'onDrop' callback in the storybook ?

yes i added an onDrop action in the story

PR #171, opened 2022-06-26
chore!: refactor value-renderer components

lpatiny:


package.json
```diff
@@ -29,6 +29,7 @@
     "prettier": "prettier --check .",
     "prettier-write": "prettier --write .",
     "serve": "vite preview",
+    "start": "vite --open",
```
targos, line 32: this is already the `dev-app` script

lpatiny: Yes I know. You would like to keep 'start' for another purpose ? We use it in NMRium as well and I thought it is practical to simply enter `npm start` to test the full application.

targos: I don't like `start` for development. In the ecosystem it is used to start the application in production

PR #174, opened 2022-06-27
DropZone Improvement

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/164

### DIFF
--- a/src/components/DropZone.tsx
+++ b/src/components/DropZone.tsx
@@ -1,24 +1,25 @@
 /** @jsxImportSource @emotion/react */
 import { css } from '@emotion/react';
 import { MouseEventHandler, useCallback, useMemo } from 'react';
-import { useDropzone } from 'react-dropzone';
+import { FileError, useDropzone } from 'react-dropzone';
 import { FaCloudUploadAlt } from 'react-icons/fa';
 
-export interface DropzoneProps {
+export interface DropZoneProps {
   color?: string;
-  onDrop: (files: File[]) => void;
+  onDrop?: <T extends File>(files: T[]) => void;
+  fileValidator?: <T extends File>(file: T) => FileError | FileError[] | null;
+  maxFiles?: number;
 }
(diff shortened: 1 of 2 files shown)


src/components/DropZone.tsx
```diff
@@ -27,28 +28,37 @@ export function DropZoneContainer(props: {
...
-      onDrop(acceptedFiles);
+    <T extends File>(acceptedFiles: T[]) => {
+      onDrop?.(acceptedFiles);
     },
     [onDrop],
   );
   const { getRootProps, getInputProps, isDragActive } = useDropzone({
+    validator: fileValidator,
     onDrop: handleOnDrop,
+    maxFiles,
```
lpatiny, line 61: Not sure we need this property. Anyway it is client side.

src/components/DropZone.tsx
```diff
@@ -27,28 +28,37 @@ export function DropZoneContainer(props: {
...
   const handleOnDrop = useCallback(
-    (acceptedFiles) => {
-      onDrop(acceptedFiles);
+    <T extends File>(acceptedFiles: T[]) => {
+      onDrop?.(acceptedFiles);
     },
     [onDrop],
   );
   const { getRootProps, getInputProps, isDragActive } = useDropzone({
+    validator: fileValidator,
```
lpatiny, line 59: Could you give an example of fileValidator in the story book ? Not sure how to use it. could we filter out to keep only 'txt' extensions ?

We have many (or should at least) have filtering capabilities in filelist-utils so I'm not sure we need to deal with this here.

lpatiny (changes requested):

src/components/DropZone.tsx
wadjih-bencheikh18: > Could you give an example of fileValidator in the story book ?

Yes okey no problem

PR #187, opened 2022-07-01
Improve measurement panel

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/182


lpatiny (changes requested):
1. Please use always an object in the callback and both should have {kind:'', measurement:''}
2. In the list of measurements please highlight the one that is currently selected (change background color for example)
3. I don't think that the last point keeping the last selected measurement for each tab is currently implemented

lpatiny (approved):
LGTM for the functionnalities. @targos could you merge the code if it looks ok to you ?

PR #188, opened 2022-07-01
Measurement info panel

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/180


lpatiny (changes requested):
I don't think the search by value is working, only key is working.

'react-inspector' in case the value is an object seems not to be implement yet.

I'm also wondering about the 2 sections, 'info' and 'meta' corresponding to the 'info' and 'meta' properties in the measurement.
There should be 2 parts (but only one query field) with a title for the 2 parts.

I just updated the https://github.com/zakodium-oss/analysis-ui-components/blob/main/public/measurements.json so that there is an info and meta property in the first measurement and one of the value is an object.

src/components/MeasurementInfoPanel.tsx
```diff
@@ -0,0 +1,105 @@
...
+      }
+
+      return (
+        <Table.Row key={key}>
+          <ValueRenderers.Text value={key} />
+          {valueCell(value)}
+        </Table.Row>
+      );
+    });
+  const infoRows = (search: string) =>
```
lpatiny, line 35: Please avoid repeating the same code twice and add a function.

lpatiny (changes requested):
Please first display the info and then meta.

Please also use a different background to be able to distinguish before info and meta.

src/components/MeasurementInfoPanel.tsx
wadjih-bencheikh18: fixed thank you

lpatiny (approved):
Functionalities LGTM. @targos could you merge the code if it looks ok ?

targos:
There's a conflict after I merged https://github.com/zakodium-oss/analysis-ui-components/pull/187

PR #195, opened 2022-07-05
docs: group panels

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/183#event-6919847232


lpatiny:
I was thinking to have something like Accordion

image

And not a completely new section like you did now

image

@wadjih-bencheikh18  Are you able to do something like Accordion ?

wadjih-bencheikh18:
> @wadjih-bencheikh18 Are you able to do something like Accordion ?

Yes no problem

PR #198, opened 2022-07-07
Loading the measurement

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/196


wadjih-bencheikh18:
@targos using relative path only work locally i think we have to update axios base url (for netlify it works without changing)

wadjih-bencheikh18:
it works now with relative url in netlify, we may try it with gh-pages

stories/measurementInfoPanel.stories.tsx
```diff
@@ -1,11 +1,12 @@
 import { Meta } from '@storybook/react';
-import { useState, useEffect } from 'react';
+import axios from 'axios';
+import { QueryClient, QueryClientProvider, useQuery } from 'react-query';
```
lpatiny, line 3: In the issue we were thinking about something like `useMeasurementExample` and the currently implementation seems qui complex to me especially that this load of the example will be reused in many places.

@targos any advice to simplify this code ? Loading the example data should be only one line in a story book I think.

lpatiny (changes requested):

stories/measurementInfoPanel.stories.tsx
targos: Yes we need to use a global decorator to wrap all stories with the query provider

wadjih-bencheikh18: > Yes we need to use a global decorator to wrap all stories with the query provider

I tired it and it cause a bug, because when we use a global decorator it will be the same QueryClient so when we change story data will stay the same as the previous one

wadjih-bencheikh18: I used the decoration now

targos:
This isn't needed anymore after switching to Ladle: https://github.com/zakodium-oss/analysis-ui-components/pull/207

PR #199, opened 2022-07-07
Improve peak panel

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/197


src/components/IRPeaksPanel.tsx
```diff
@@ -1,31 +1,117 @@
...
+  flexRender,
+  SortingState,
+  getSortedRowModel,
+} from '@tanstack/react-table';
+import { useState } from 'react';
 
-import { Table } from './Table';
 import { IRPeak } from './context/data/DataState';
 
+type IRPeakPanelPreferences = Record<
```
lpatiny, line 13: preferences could have more properties so better that is it an object and one of the property is 'columns'.

Now concerning the columns is it better to have an object or rather an array containing key, etc. After in the code you need to make some maps and it would probably be more intuitive on an array.

src/components/IRPeaksPanel.tsx
```diff
@@ -1,31 +1,117 @@
...
+    }),
   );
+
+  const [sorting, setSorting] = useState<SortingState>([]);
+  const table = useReactTable({
+    data: peaks,
+    columns,
+    state: {
+      sorting,
+      columnVisibility: Object.entries(preferences).reduce(
```
lpatiny, line 58: This code looks quite complex. I hope that if the preferences.columns is an array is could make it simpler.

Anyway if a `map` is not possible please don't use `reduce` and use normal loops in a method called `getColumnVisibility`

lpatiny (changes requested):

src/components/IRPeaksPanel.tsx
wadjih-bencheikh18: So you perfer that we use an array containing key?

src/components/IRPeaksPanel.tsx
wadjih-bencheikh18: fixed

src/components/IRPeaksPanel.tsx
wadjih-bencheikh18: i used the suggested type in the issue "array containing key"

src/components/IRPeaksPanel.tsx
```diff
@@ -1,31 +1,117 @@
...
+                      header.column.columnDef.header,
+                      header.getContext(),
+                    )}
+                    {{
+                      asc: ' 🔼',
+                      desc: ' 🔽',
+                    }[header.column.getIsSorted() as string] ?? null}
+                  </div>
+                )}
+              </th>
```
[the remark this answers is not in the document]
wadjih-bencheikh18: i implemented it in value.renderers.Title as a param `sorted`

lpatiny (approved):
LGTM

@targos could you check and merge if ok ? I'm not familiar with react-table.

src/components/value-renderers/Title.tsx
```diff
@@ -6,17 +6,24 @@ import { ValueRenderersProps } from '.';
...
+  sorted?: 'asc' | 'desc' | false;
 }
 
 const styles = {
   title: css({
     fontWeight: 'bold',
   }),
 };
-export function Title({ value, ...other }: TitleProps) {
+export function Title({ value, sorted = false, ...other }: TitleProps) {
```
stropitek, line 17: It does not really make sense to me that the `sorted` prop is added to a generic `Title` component.

I think you should have another component specifically for rendering table headers. It can reuse the `Title` component if needed.

src/components/IRPeaksPanel.tsx
```diff
@@ -1,31 +1,94 @@
...
+
+import { Table, ValueRenderers } from '..';
 
-import { Table } from './Table';
 import { IRPeak } from './context/data/DataState';
 
+interface ColumnPreferences {
+  visible?: boolean;
+  format?: (val: number | string) => string | number;
+  jpath: keyof IRPeak;
```
stropitek, line 17: I think you can keep the name `accessorKey` since `jpath` to me implies you can access deep properties, which react-table does not allow (you would have to define an accessor function instead).

src/components/value-renderers/Title.tsx
wadjih-bencheikh18: table header added thanks

wadjih-bencheikh18:
@stropitek  all story cases added , thank you!

stropitek:
LGTM

> What is the added value of having `visible: false` versus just not include the item in the array?

@lpatiny maybe more a question for you since you have more context about what underlying data those props will be derived from.

src/components/value-renderers/Header.tsx
```diff
@@ -1,13 +1,12 @@
-import { Text } from './Text';
-import { TitleProps } from './Title';
+import { Title, TitleProps } from './Title';
 
 interface HeaderProps extends TitleProps {
   sorted?: 'asc' | 'desc' | false;
 }
 export function Header({ sorted = false, ...other }: HeaderProps) {
   return (
     <div>
```
targos, line 8: Maybe this should be a `<th>` ?
Also I'm not sure `value-renderers` is the best place to put table-specific components.

wadjih-bencheikh18: `<td/>` is added in Row component

targos: But this is a header, so we should have `<th>` in the DOM, not `<td>`

wadjih-bencheikh18: > Also I'm not sure `value-renderers` is the best place to put table-specific components.

could you please suggest the right place?

lpatiny:
> LGTM
> 
> > What is the added value of having `visible: false` versus just not include the item in the array?
> 
> @lpatiny maybe more a question for you since you have more context about what underlying data those props will be derived from.

The goal is that the enduser from a future preference panel can decide which columns are visible or not.

PR #201, opened 2022-07-11
Signal processing panel

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/200


lpatiny:
I would need the possibility to have no filter at all so by default it should be empty (only the + sign)

I would like to see in the storybook the result of the callback. Could you add a component displaying the JSON ? This is quite a general comment and it would be nice in the storybook to see directly what is sent to the callback.

Can you take into account the default value:
```json
  "fromIndex": {
      "default": 0,
```

Would be nice to display a tip if available, maybe under the input field for example. You should just not care to the layout / design for now.

```json
  "to": {
      "description": "Last value for xyIntegration in the X scale",
```

Some options are of type 'enum' and there should be a 'select' input field with the allowed values
```
  "name": {
      "enum": [
          "normed"
      ],
      "type": "string"
  },
  "options": {
      "properties": {
          "algorithm": {
              "default": "'absolute'",
              "description": "algorithm can be 'sum' 'max' or 'absolute'",
              "enum": [
                  "absolute",
                  "max",
                  "sum"
              ],
```

We need to be sure we are able to reload an existing list of filters with their configuration. Maybe this could be a second storybook.

The schema (FilterXYSchema.json) should not be copied in this project but directly be retrieve from the corresponding package that will become a dependency of this package.

wadjih-bencheikh18:
> I would need the possibility to have no filter at all so by default it should be empty (only the + sign)

image

here the drop down menu is not a selected filter so in this example there is 3 selected filters
if you see that we should consider it as selected i can make modification

wadjih-bencheikh18:
@lpatiny Could you please check?

PR #213, opened 2022-09-15
feat: drag and drop measurements in App page

stropitek:
- feat: load data from drag and drop in app page
- feat: load selected measurement


src/app/App.tsx
```diff
@@ -30,107 +64,134 @@ export default function App() {
...
-                <FaTabletAlt />
-              </Toolbar.Item>
-            </Toolbar>
-          </Header>
-        </div>
+    <DropZoneContainer
+      onDrop={(files) => {
+        void loadFiles(files, dispatch);
+      }}
+    >
```
targos, line 72: Can you put this inside `RootLayout`? Otherwise it's not part of the shadow dom.

src/components/DropZone.tsx
```diff
@@ -76,6 +76,7 @@ function DropZoneContent(
         height: 100%;
         width: 100%;
         ${isDragActive || !isContainer ? `border: 2px dashed ${color};` : null}
+        box-sizing: border-box;
```
targos, line 79: This change isn't needed if you put the component in the shadow dom

src/app/context/appState.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { assertUnreachable } from '../../utils/assert';
+
+import { getFirstMeasurement, getMeasurementOrFail } from './data.helpers';
+
+export interface AppState {
+  data: DataState;
+  isLoading: boolean;
+  view: {
+    selectedMeasurement?: string;
+  };
```
targos, line 25: It's fine as a first iteration, but I think that the goal is to have one selected measurement per "tab" (i.e. if you select an IR spectrum X, then go to the Raman tab and select Raman spectrum Y, then go back to the IR tab, it should remember that spectrum X is selected there.

PR #220, opened 2022-09-22
chore: refactor file structure

stropitek:


lpatiny (commented):
I would create a 'panels' folder because we will have a lot of them.

It is also not clear for me what should be in 'app' and what should be in 'components'. I thought what is reusable in many application should be in apps and MeasurementExplorer, MeasurementInfoPanel, MeasurementPlot and MeasurementsPanel are expected to be reused.

PR #227, opened 2022-09-28
reimplement modal with a dialog element

stropitek:
- feat: add width and height props to ConfirmModal
- feat!: implement Modal using dialog


stropitek:
I don't think we can use this because I just noticed the previous implementation is placing the modal with respect to the "root" element, but using dialog prevent us from doing that :\

targos:
There's a type error. Can I review already?

PR #238, opened 2022-10-08
feat: save as ium

wadjih-bencheikh18:
Ref: https://github.com/zakodium-oss/analysis-ui-components/issues/216


lpatiny:
Functionality looks ok to me.

src/app/context/appState.tsx
```diff
@@ -157,7 +162,23 @@ function actionHandler(draft: Draft<AppState>, action: AppStateAction) {
       draft.isLoading = false;
       return;
     }
+    case 'SAVE_AS_IUM': {
+      saveAsIum(draft, action.payload?.filename, action.payload?.spaceIndent);
+      return;
+    }
```
stropitek, line 168: You should not have side effects in a reducer.

Actions should only update the state.

src/app/App.tsx
```diff
@@ -69,6 +70,9 @@ function DropZoneArea() {
     { id: '1h,1h', title: '1H,1H', content: 'Hello, World! [c]' },
     { id: '1h,13c', title: '1H,13C', content: 'Hello, World! [d]' },
   ];
+  function saveHandler() {
+    dispatch({ type: 'SAVE_AS_IUM' });
+  }
```
stropitek, line 75: Please implement the download logic here, not in the reducer. Please implement the download without the `file-saver` library.

It's relatively easy to do, something like this:

```ts
export function download(data: string) {
  const link = document.createElement('a');

  const blob = new Blob([data]);
  const url = URL.createObjectURL(blob);


  link.setAttribute('href', url);
  link.setAttribute('download', name);
  link.style.visibility = 'hidden';
  document.body.appendChild(link);
  link.click();
  document.body.removeChild(link);
  URL.revokeObjectURL(url);
}
```

wadjih-bencheikh18: fixed thank you

stropitek:
Save LGTM

Loading the `.ium` is not yet implemented

lpatiny:
@stropitek Could we merge this PR ? Loading the .ium should be part of another PR.

PR #242, opened 2022-10-11
feat: many tab scroll

wadjih-bencheikh18:
closes https://github.com/zakodium-oss/analysis-ui-components/issues/215
I implemented a basic solution which is adding scroll on overflow.

### DIFF
--- a/src/components/Tabs.tsx
+++ b/src/components/Tabs.tsx
@@ -1,6 +1,6 @@
 /** @jsxImportSource @emotion/react */
 import { css } from '@emotion/react';
-import type { ReactNode } from 'react';
+import { ReactNode, useMemo, useRef } from 'react';
 
 import { TabsProvider, useTabsContext } from './context/TabsContext';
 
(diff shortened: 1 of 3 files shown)


src/components/Tabs.tsx
```diff
@@ -123,15 +130,24 @@ function TabsHorizontal<T extends string = string>(
...
+    <div
+      style={{ display: 'flex', flexDirection: 'column', overflowX: 'hidden' }}
+    >
       <div
+        ref={divRef}
         style={{
           display: 'flex',
           flexDirection: 'row',
-          height: 38,
+          height: hasScroll ? 60 : undefined,
```
targos, line 148: This doesn't seem necessary? In the preview, the height is not set and it still seems to work for me:
image

stropitek:
It would be nice if the scrollbar was not that big. Modern css rules allow you to change that.

https://blog.mayank.co/better-scrolling-through-modern-css

wadjih-bencheikh18:
> It would be nice if the scrollbar was not that big.

scrollbar style changed

PR #254, opened 2022-10-16
test: added biologic test back

santimirandarp:
Seems to be accepting the biologic-converter data, but it errors out :

```
Uncaught TypeError: value is undefined
convertString react-plot.js:4246
    convertValue react-plot.js:4251
    usePosition react-plot.js:4157
    Line react-plot.js:4385
    React 11
    workLoop scheduler.development.js:266
    flushWork scheduler.development.js:239
    performWorkUntilDeadline scheduler.development.js:533
    js scheduler.development.js:571
    js
```

### DIFF
--- a/src/app/data/loaders/biologicLoader.ts
+++ b/src/app/data/loaders/biologicLoader.ts
@@ -20,13 +20,13 @@ export const biologicLoader: Loader = async function biologicLoader(
       title: '',
     };
     if (mpr !== undefined) {
-      prepare.meta = { settings: mpr.settings };
+      prepare.meta = { ...mpr.settings.variables };
       prepare.data = [{ variables: mpr.data.variables }];
     } else if (mpt !== undefined) {
-      prepare.meta = { settings: mpt.settings };
+      prepare.meta = { ...mpt.settings.variables };
       prepare.data = [{ variables: mpt.data.variables }];
     } else if (mps !== undefined) {
-      prepare.meta = { settings: mps.settings };
+      prepare.meta = { ...mps.settings.variables };
     }
     measurements.iv.entries.push(prepare as MeasurementBase);
   }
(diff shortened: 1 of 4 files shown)


src/app/data/loaders/biologicLoader.ts
```diff
@@ -3,23 +3,32 @@ import { convert } from 'biologic-converter';
...
       filename: '',
-      path: result.dir,
+      path: dir,
       info: {},
       title: '',
-      data: [],
-    });
+    };
+    if (mpr !== undefined) {
+      prepare.meta = { settings: mpr.settings, log: mpr?.log };
```
lpatiny, line 23: For meta it is better if it is flat and the log is not that important.
Something like:
`prepare.meta = {...mpr.settings)`
We will display those values and it will be searchable in the following pannel: https://zakodium-oss.github.io/analysis-ui-components/stories/?story=measurements--panels--measurement-info-panel

src/app/data/loaders/biologicLoader.ts
```diff
@@ -3,23 +3,32 @@ import { convert } from 'biologic-converter';
...
+      prepare.data = [{ variables: mpr.data.variables }];
+    } else if (mpt !== undefined) {
+      prepare.meta = { settings: mpt.settings, log: mpt.log };
+      prepare.data = [{ variables: mpt.data.variables }];
+    } else if (mps !== undefined) {
+      prepare.meta = { settings: mps.settings, log: mps.log };
```
lpatiny, line 29: Please flatten everywhere.

lpatiny (commented):
Seems it work correcly, great ! I'm just wondering which variable we should display by default.

Ewe versus time looks like a good choice.


image

src/app/data/loaders/biologicLoader.ts
santimirandarp: It should be done now. Are you running `npm run dev-app` as well? I wonder if why would we see different output, could be some update in between.

PR #256, opened 2022-10-19
feat: add responsive chart option

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/249


wadjih-bencheikh18:
we need to solve https://github.com/zakodium-oss/react-d3-utils/issues/26 in order to fix types error

lpatiny:
@wadjih-bencheikh18 is it expected to already work ? Because currently the spectrum do not take the full space yet

image

wadjih-bencheikh18:
> is it expected to already work ?

I didn't change the app I updated only the story

stropitek:
There are a few issues when resizing the window / split pane

targos:
No, it is not expected.

targos:
But I think it's a known issue, unrelated to this PR

wadjih-bencheikh18:
> But I think it's a known issue, unrelated to this PR

i will try to figure out why this is happening

wadjih-bencheikh18:
The problem is with `ResponsiveChart` because this latter give to his child static width so we're in evil child splitpane case,

lpatiny:
There is also the issue that the chart does not take the full height. It should not only take the full width but also the full height.

targos:
> The problem is with `ResponsiveChart` because this latter give to his child static width so we're in evil child splitpane case,

Yes exactly. I think I know how to fix it and will wait until this pr is merged.

wadjih-bencheikh18:
> Yes exactly.

There's also another bug in responsive chart when using percentage height

targos:
It has a height of 0 for me: 
image

src/app/components/MeasurementPlot.tsx
```diff
@@ -104,36 +107,51 @@ function MeasurementComponent(props: MeasurementPlotProps) {
...
+    <div
+      css={css`
+        user-drag: none;
+        -webkit-user-drag: none;
+        user-select: none;
+        -moz-user-select: none;
+        -webkit-user-select: none;
+        -ms-user-select: none;
+      `}
+    >
```
targos, line 119: This div doesn't have a height. That's why the responsive chart takes 0px.

wadjih-bencheikh18: fixed, is it expected that the plot keep stretching

targos: > is it expected that the plot keep stretching

no

wadjih-bencheikh18: bug fixed

src/app/components/MeasurementExplorer.tsx
```diff
@@ -37,7 +37,15 @@ export function MeasurementExplorer(props: MeasurementExplorerProps) {
...
   return (
-    <div style={{ width, height }}>
+    <div
+      style={{ width, height }}
+      css={css`
+        display: flex;
+        flex-direction: column;
+        gap: 40px;
+        justify-content: space-around;
+      `}
```
stropitek, line 47: Can you put all the styles either in `style` or in `css`, instead of having a mix of both?

stropitek: Also, do you have an explanation for the bug and why this resolves it?

wadjih-bencheikh18: > do you have an explanation for the bug and why this resolves it?

MeasurementExplorer contain MeasurementPlot + select inputs (where we can change variables)
MeasurementPlot was taking 100% of it, but inputs have also a height so each time the page is rendring the plot change height ..etc (so it keeps stretching)

as a solution i used flex-col to divide MeasurementExplorer

wadjih-bencheikh18: > Can you put all the styles either in `style` or in `css`, instead of having a mix of both?

It will be  
```
width: ${typeof width === 'number' ? `${width}px` : width};
height: ${typeof height === 'number' ? `${height}px` : height};
```
in css

stropitek: Alright you can leave it like it is then

stropitek:
image


The tests are failing because the controls in MeasurementExplorer seem to be rendered outside of its container. This is why in the tests the button cannot be clicked (see in the screenshot how the control bar is almost invisible)

This can also be seen when looking at the story for MeasurementExplorer

wadjih-bencheikh18:
> the controls in MeasurementExplorer seem to be rendered outside of its container.

fixed, it was because of justify-content

PR #267, opened 2022-10-21
feat: load ium state

wadjih-bencheikh18:
closes https://github.com/zakodium-oss/analysis-ui-components/issues/216


lpatiny (approved):
Functionality is ok for me.

stropitek (changes requested):
Loading an `.ium` file should be a separate path than loading other types of files, and it's not a "loader".

You should check if the file collection contains a unique nmrium file. If yes replace the current app state with the new state. If not use the loaders to add measurements.

src/app/App.tsx
```diff
@@ -102,7 +104,15 @@ function DropZoneArea() {
     <RootLayout>
       <DropZoneContainer
         onDrop={(files) => {
-          void loadFiles(files, dispatch);
+          if (
+            isEqual(appState, getEmptyAppState()) &&
```
stropitek, line 108: I think you can remove this check. We want to replace the state entirely in that case (later we will surely implement a confirmation since it's destructive)

src/app/data/append.ts
```diff
@@ -39,16 +39,16 @@ export async function append(
   return { logs: [], dataState: nextDataState };
 }
 
-export async function loadMeasurements(
+export async function loadData(
```
stropitek, line 42: You can keep the original name `loadMeasurements` as it is more expressive

src/app/data/loaders/iumLoader.ts
```diff
@@ -0,0 +1,25 @@
...
+import {
+  getEmptyMeasurements,
+  Loader,
+  Measurements,
+  mergeMeasurements,
+} from '../DataState';
+
+import { assert } from '@/utils/assert';
+
+export const iumLoader: Loader = async function iumLoader(
```
stropitek, line 12: You can remove the ium loader entirely. An `ium` file should never be handled unless it's the only file that was dropped. You can think of an `ium` file in our application as of a `psd` file in photoshop, if the analogy helps.

src/app/App.tsx
```diff
@@ -102,7 +104,15 @@ function DropZoneArea() {
...
+            files.length === 1 &&
+            files[0].name.match(/\.ium$/i)
+          ) {
+            void loadIUM(files[0], dispatch);
+          } else {
+            void loadFiles(files, dispatch);
+          }
```
stropitek, line 115: You can move this function to the `load.ts` file and import it from there

src/app/data/loaders/iumLoader.ts
wadjih-bencheikh18: fixed

stropitek (commented):
LGTM after my last comment is resolved

src/app/context/load.ts
```diff
@@ -28,18 +28,32 @@ const options = {
...
-        ? files
-        : await fileCollectionFromFiles(files);
-    const measurements = await loadMeasurements(fileCollection, options);
-    dispatch({ type: 'ADD_MEASUREMENTS', payload: measurements });
+    if (files[0] && !files[1] && files[0].name.match(/\.ium$/i)) {
+      const data = await files[0].text();
+      const appState = JSON.parse(data);
+      const payload: AppState = {
+        ...appState,
+        isLoading: true,
```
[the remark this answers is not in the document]
wadjih-bencheikh18: fixed, thank you

stropitek (approved):

PR #268, opened 2022-10-21
feat: select measurement on drop

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/analysis-ui-components/issues/250

### DIFF
--- a/src/app/context/appState.tsx
+++ b/src/app/context/appState.tsx
@@ -107,7 +107,10 @@ function actionHandler(draft: Draft<AppState>, action: AppStateAction) {
             kind,
           );
           if (measurement) {
-            draft.view.selectedMeasurements[kind] = measurement.measurement.id;
+            const id = measurement.measurement.id;
+            draft.view.selectedMeasurements[kind] = id;
+            draft.view.currentMeasurement = { id, kind };
+            break;
             // draft.view.currentMeasurement = {
             //   kind,
             //   id: measurement.measurement.id,
--- a/src/app/context/appState.tsx
+++ b/src/app/context/appState.tsx
@@ -109,8 +109,9 @@ function actionHandler(draft: Draft<AppState>, action: AppStateAction) {
           if (measurement) {
             const id = measurement.measurement.id;
             draft.view.selectedMeasurements[kind] = id;
-            draft.view.currentMeasurement = { id, kind };
-            break;
+            if (draft.view.currentMeasurement === undefined) {
+              draft.view.currentMeasurement = { id, kind };
+            }
             // draft.view.currentMeasurement = {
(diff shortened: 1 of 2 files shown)


lpatiny (approved):
Functionality looks ok to me.

src/app/context/load.ts
```diff
@@ -32,6 +33,20 @@ export async function loadFiles(
...
+        kind = k as MeasurementKind;
+        break;
+      }
+    }
+    if (kind) {
+      dispatch({
+        type: 'SELECT_MEASUREMENT',
+        payload: { id: measurements[kind].entries[0].id, kind },
+      });
+    }
```
targos, line 49: I think this should be done inside the ADD_MEASUREMENTS action, and only if there is nothing selected yet. @stropitek ?

stropitek: Yes I think this logic can be moved to the ADD_MEASUREMENTS action

wadjih-bencheikh18: Fixed, thanks.

lpatiny:
Auto select measurement should only be done if no measurement is currently selected.

In the current version there is a weird behaviour. If you drag / drop ir2, mass and uv there is auto select of ir2. This is ok.

However after if you add one more 'ir' suddently there is the UV that is selected.



Using as rule that you auto-select only if nothing is selected you will not have this issue.

wadjih-bencheikh18:
> Auto select measurement should only be done if no measurement is currently selected.

Fixed, could you please try now.

lpatiny:
Yes the functionality is ok now.

src/app/context/appState.tsx
```diff
@@ -107,7 +107,11 @@ function actionHandler(draft: Draft<AppState>, action: AppStateAction) {
             kind,
           );
           if (measurement) {
-            draft.view.selectedMeasurements[kind] = measurement.measurement.id;
+            const id = measurement.measurement.id;
+            draft.view.selectedMeasurements[kind] = id;
+            if (draft.view.currentMeasurement === undefined) {
+              draft.view.currentMeasurement = { id, kind };
+            }
```
targos, line 114: This doesn't sound right. It means we can have a different current measurement to the selected measurement.

stropitek: I made this part of the state in https://github.com/zakodium-oss/analysis-ui-components/commit/2df2cc06780a2fb1459d67388fda7495423a1232

I think we can change `currentMeasurement` to be `displayedKind`. The measurement displayed would be the `selectedMeasurement[displayedKind]`

@targos I suggest we create a new issue for that since the functionality for this PR seems to be ok

PR #283, opened 2022-10-25
DropDown menu

Sebastien-Ahkrin:
Parts of https://github.com/zakodium-oss/analysis-ui-components/issues/260


src/components/menu/MenuItems.tsx
```diff
@@ -0,0 +1,152 @@
...
+    </ClassNames>
+  );
+}
+
+function Item<T>(props: ItemProps<T>) {
+  const { option, onSelect } = props;
+  const isDivider = option.type === 'divider';
+
+  return (
+    <Menu.Item disabled={!isDivider && option.disabled}>
```
stropitek, line 91: The divider should not render a `Menu.Item`. `Menu.Item` is for elements that can be active and selected with the keyboard (so only options of type `option`)

src/components/menu/MenuItems.tsx
```diff
@@ -0,0 +1,152 @@
...
+  );
+}
+
+function ItemOption<T>(props: ItemOptionProps<T>) {
+  const { onSelect, option, active } = props;
+  const { disabled } = option;
+
+  return (
+    <ClassNames>
+      {({ css }) => (
```
stropitek, line 111: There is a simpler way to pass css. See for example https://github.com/zakodium-oss/analysis-ui-components/blob/main/src/components/value-renderers/Color.tsx

The first line of the file is important!

src/components/menu/MenuItems.tsx
```diff
@@ -0,0 +1,152 @@
...
+        <div
+          onClick={() => onSelect(option)}
+          className={css([
+            {
+              ...styles.item,
+              display: 'flex',
```
stropitek, line 117: did you not use grid (was suggested in the issue) for a reason?

stories/components/menuitems.stories.tsx
```diff
@@ -0,0 +1,71 @@
+import { Menu } from '@headlessui/react';
+import { useMemo } from 'react';
+import { FaMeteor } from 'react-icons/fa';
+
+import { MenuItems, MenuOptions } from '@/components/menu/MenuItems';
+
+export default {
+  title: 'Components / MenuItems',
```
stropitek, line 8: You can move these stories and the "Dropdown" stories to a common "Components / DropdownMenu" folder.

src/components/menu/MenuItems.tsx
Sebastien-Ahkrin: no, just a personnal preference, i will use grid on the next push :)

Sebastien-Ahkrin: what, how can i reproduce this bug ?

stropitek:
image

The labels should be aligned vertically, like in [this screenshot](https://user-images.githubusercontent.com/2352663/197483598-331a0702-0b3f-476f-8f1a-e96d4a6b819e.png)

src/components/menu/MenuItems.tsx
```diff
@@ -107,8 +107,8 @@ function ItemOption<T>(props: ItemOptionProps<T>) {
       css={css([
         {
           ...styles.item,
-          display: 'flex',
-          flexDirection: 'row',
+          display: 'grid',
```
targos, line 110: It's not the option that should be a grid. It's the entire menu.

lpatiny:
When you open the drop down menu and then click somewhere else I would expect the menu closes: https://deploy-preview-283--analysis-ui-components.netlify.app/stories/?story=components--dropdownmenu--dropdown

Also the context menu seems to crash: https://deploy-preview-283--analysis-ui-components.netlify.app/stories/?story=components--dropdownmenu--context-dropdown

Sebastien-Ahkrin:
> When you open the drop down menu and then click somewhere else I would expect the menu closes: https://deploy-preview-283--analysis-ui-components.netlify.app/stories/?story=components--dropdownmenu--dropdown
> 
> Also the context menu seems to crash: https://deploy-preview-283--analysis-ui-components.netlify.app/stories/?story=components--dropdownmenu--context-dropdown

yeah sorry @lpatiny i'm still working on it, i mark the Pull request as draft sorry !

lpatiny:
I'm eager that we replace in NMRium the current drop down by this new one that looks soo much better !!!

I have a specific case for which I would like a different icon color. Something like:

image

Does the current implementation allows this ?

targos:
> Does the current implementation allows this ?

Yes, the icon can be any React element

PR #284, opened 2022-10-25
refactor[WIP]: show axes as "name (unit)"

santimirandarp:
* Issue
The current MeasurementExplorer shows the variables:

Screenshot from 2022-10-25 16-02-23

* Changes

This is how it currently looks like (still didn't change the test case.)

Screenshot from 2022-10-25 16-01-33

There is another issue, if the `<Annotations>` is commented out, we can switch bw variables and see all the different plots, but otherwise it crashes. I couldn't find out the specific problem.


santimirandarp:
> Indeed displaying the axes labels + units is much better.
> 
> For the legend story what is the message when it crashes ?

Thanks. I am adding the rest. The merge conflict is strange, I think we can safely keep the dev changes.

I scoped the error down to [this lines in react-plot](https://github.com/zakodium-oss/react-plot/blob/612a5aa6ba944638062e184dba9b4712583e1986/src/hooks/useCrossHair.tsx#L67), I may be fully wrong though. 

This is the error (dissapears if I comment out crossHair annotations.):

```
The above error occurred in the <Line> component:

Line@http://localhost:5174/node_modules/.vite/deps/react-plot.js?v=00e82732:4384:123
Annotations@http://localhost:5174/node_modules/.vite/deps/react-plot.js?v=00e82732:4540:73
g
g
svg
div
LegendProvider@http://localhost:5174/node_modules/.vite/deps/react-plot.js?v=00e82732:3625:44
Plot@http://localhost:5174/node_modules/.vite/deps/react-plot.js?v=00e82732:7112:180
MeasurementComponent@http://localhost:5174/src/app/components/MeasurementPlot.tsx?t=1667152342949:53:7
NestedContextProvider@http://localhost:5174/node_modules/.vite/deps/react-plot.js?v=00e82732:3970:37
NestedContextProvider@http://localhost:5174/node_modules/.vite/deps/react-plot.js?v=00e82732:3970:37
NestedContextProvider@http://localhost:5174/node_modules/.vite/deps/react-plot.js?v=00e82732:3970:37
NestedContextProvider@http://localhost:5174/node_modules/.vite/deps/react-plot.js?v=00e82732:3970:37
PlotController@http://localhost:5174/node_modules/.vite/deps/react-plot.js?v=00e82732:4120:28
MeasurementPlot
```

Same error occurs in the Text component. @lpatiny

lpatiny:
What is the state of this PR ?

Because I see that in the main branch we have already the labels in the menu

image

santimirandarp:
> What is the state of this PR ?
> 
> Because I see that in the main branch we have already the labels in the menu
> 
> image

Yes, but those labels come from the old biologic format, I think the dep was not updated maybe. It returns one letter variables now, a-zA-Z, so the explorer would show single letters as in the other Measurements, if you try any of them (I believe.)

The issue is that whenever a flag is used, the crossHair Annotation breaks, but I am not sure whether the error is there, or in the values passed to it.

```
Uncaught TypeError: value is undefined
    convertString hooks.ts:186
    convertValue hooks.ts:193
    usePosition hooks.ts:25
    Line Line.tsx:31
```
If the crossHair is commented though, everything seems to work, so I may add a separate issue for it.

PR #286, opened 2022-10-27
Many measurmenet

wadjih-bencheikh18:
we can now select many measurements, i didn't update `measurementExplorer` to accept many measurement so i used the first one

in order to update it, I need more details about dataIndex and X, Y Variables values

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -183,8 +183,8 @@ function DropZoneArea() {
                   }}
                 >
                   <div style={{ padding: 5 }}>
-                    {measurement ? (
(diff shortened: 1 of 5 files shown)


targos:
Is there an issue related to this? I think it's too early to implement. For example https://github.com/zakodium-oss/analysis-ui-components/issues/280 and https://github.com/zakodium-oss/analysis-ui-components/issues/269 should have been done first.

wadjih-bencheikh18:
> I think it's too early to implement.

okey i will reopen the PR once it's the time

PR #290, opened 2022-10-29
feat: cdf loader for gcms added

josoriom:


src/app/data/loaders/cdfLoader.ts
```diff
@@ -0,0 +1,119 @@
...
+  MeasurementKind,
+  Loader,
+  Measurements,
+  getEmptyMeasurements,
+} from '../DataState';
+
+export const cdfLoader: Loader = async function cdfLoader(
+  fileCollection: FileCollection,
+) {
+  const newMeasurements: Measurements = getEmptyMeasurements();
```
lpatiny, line 17: Please add a comment with the link to the github code from which you was inspired.

src/app/data/loaders/cdfLoader.ts
```diff
@@ -0,0 +1,119 @@
...
+      const tic = reader.getDataVariable('total_intensity');
+      const pointCount = reader.getDataVariable('point_count');
+      const massValues = reader.getDataVariable('mass_values');
+      const intensityValues = reader.getDataVariable('intensity_values');
+      addMeta(reader, reader.globalAttributes);
+
+      let index = 0;
+      for (let i = 0; i < pointCount.length; i++) {
+        const size = pointCount[i];
+        const mass = new Array(size);
```
lpatiny, line 39: Please use typed array Float64

src/app/data/loaders/cdfLoader.ts
```diff
@@ -0,0 +1,119 @@
...
+        if (kind) {
+          newMeasurements[kind].entries.push({
+            id: v4(),
+            meta: reader.header.meta,
+            filename: file.name,
+            path: file.relativePath || '',
+            info: reader.getAttribute('experiment_title'),
+            title: reader.getAttribute('experiment_title'),
+            data: normalizeChromatogram({
+              time: time[i],
```
lpatiny, line 54: The data should be an array, one entry per time.

In the info:
https://github.com/zakodium-oss/analysis-ui-components/blob/43adbe8737760eab6db24acca0ed53a588716a4d/src/app/data/MeasurementBase.ts#L20 we can have tic and bpc.

lpatiny (commented):
This experiment should appear like one measurement. This means that in the tab on the top right there should be only one line

image

What we expect after the parsing in this case is to have many 'dataincdex'

image

One entry by time. 

Currently the component that displays the data is a 'debug component' that allows to inspect everything that is in the measurement but after we will have specific component for a specific measurement. So in this case a specific GC / MS or LC / MS component.

lpatiny:
You can not test it directly in the preview



Or test it locally using `npm run app-dev`. You can either drag and drop the agilent netcdf file or also click on the top link.

src/app/data/loaders/cdfLoader.ts
josoriom: So, it's ok?

lpatiny: You store indeed in the right place but for one GC/MS there should be only one measurement entries and each 'time' is an element in the data array.

https://github.com/zakodium-oss/analysis-ui-components/blob/43adbe8737760eab6db24acca0ed53a588716a4d/src/app/data/MeasurementBase.ts#L18-L22

lpatiny:
I still had to do many changes
1. If it is an array let's try to name systematically the variable in plural (masses, intensities)
2. If we have an array or array like in this case I named them allMasses, allIntensities
3. I removed the bpc because it is a derived data. We have a way to have a 'postProcessor' and we will do

PR #295, opened 2022-10-29
feat: add info panel to app

wadjih-bencheikh18:
closes https://github.com/zakodium-oss/analysis-ui-components/issues/253

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -10,7 +10,11 @@ import {
   FaSave,
 } from 'react-icons/fa';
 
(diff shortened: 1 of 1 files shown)


lpatiny:
Functionality looks ok to me. However we need to discuss what will happen if many measurements are selected ?

https://github.com/zakodium-oss/analysis-ui-components/pull/286

I guess in this case we should for now display a message like 'Information is only available when one (and only one) measurement is selected'

The same message should be displayed if no measurement is selected.

PR #296, opened 2022-10-29
feat: save using shortcuts

wadjih-bencheikh18:
closes https://github.com/zakodium-oss/analysis-ui-components/issues/279


lpatiny (approved):
Functionality is ok for me.

PR #302, opened 2022-10-31
refactor: use hash params instead of search params

targos:
Closes: https://github.com/zakodium-oss/analysis-ui-components/issues/258

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,10 +44,10 @@ When create a new loader you also need to specify that it should be used when pr
 
 ## Testing the demo application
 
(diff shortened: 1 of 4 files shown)


targos:
This has an interesting side-effect. Now when you click on the example links, the data are appended to the existing state instead of reloading everything.

stropitek (approved):
Nice trick using `useSyncExternalStore` 👍🏻

lpatiny:
I like this improvement to append and not clear the state !

PR #303, opened 2022-11-01
Mass plot

wadjih-bencheikh18:
closes https://github.com/zakodium-oss/analysis-ui-components/issues/255


wadjih-bencheikh18:
I'm not sure why annotations are not displayed

lpatiny:
The tool bar that is on the top:

image

is part of the 'debug' component to display general measurements and should disappear when we display 'mass' spectra (the new component that you are adding).

It is possible that the peak problem is my fault. I added in the main branch a new 'mass' example. Please rebase your branch and test with the second entry of the mass example. Should look like that:

image

lpatiny (approved):
For me this PR may be merge. I will try to take it over to deal with calculations and add the annotations correctly when zooming but this will be in another PR.

PR #307, opened 2022-11-01
feat: use smaller base font size in components

stropitek:


stropitek:
@lpatiny @targos 

What do you think of the new size? It mainly affects Tabs, Accordion, and anything with tables

I wanted to use a relatively small base size (12px) since generally the UI will have to be very compact.

stropitek:
The size in App is broken, you can just look at the storybook for now

stropitek:
Fixed the issue in App

targos:
I don't like the smaller icons.

stropitek:
> I don't like the smaller icons

I agree, I did not notice them before. Changed that, but they are still slightly smaller than before.

targos (approved):

lpatiny:
Nice improvement ! For scientific applications in which we need soo many options it is better to have this compact design (I know I have new glasses ;).

PR #309, opened 2022-11-02
fix: plot mass with labels

lpatiny:


lpatiny:
@wadjih-bencheikh18 Could you take over this PR ?

Need to remove all the xAxis, yAxis if not necessary (everywhere). By default it is x and y and it should be ok.

Please also check the test cases

src/app/components/MeasurementMassPlot.tsx
```diff
@@ -94,19 +92,19 @@ function MassComponent(props: MeasurementPlotProps) {
         <LineSeries
           data={profile}
           lineStyle={{ stroke: 'green' }}
-          xAxis={xAxis}
-          yAxis={yAxis}
+          xAxis="x"
+          yAxis="y"
```
targos, line 96: xAxis is "x" by default, and yAxis is "y" by default. Please remove these props everywhere unless a different value is passed.

wadjih-bencheikh18:
> Need to remove all the xAxis, yAxis

I removed all axisID as mentioned, and updated tests

src/app/components/MeasurementMassPlot.tsx
wadjih-bencheikh18: fixed

PR #319, opened 2022-11-04
refactor(MeasurementExplorer): dropdown items

santimirandarp:
Display some information in the dropdown (it uses the variable id, label and units).  The id is for avoiding confusion between variables with the same label (say both are "Transmittance").

But should we lowercase the label, uppercasing the first letter, or leave them as they appear in the files?

**Before**

Screenshot from 2022-11-04 20-26-11

**After changes**

Screenshot from 2022-11-04 20-19-27

### DIFF
--- a/src/app/components/MeasurementExplorer.tsx
+++ b/src/app/components/MeasurementExplorer.tsx
@@ -15,6 +15,8 @@ interface ExplorerInfo {
   yVariableName: string;
   flipHorizontalAxis: boolean;
 }
+type XOrY = keyof Pick<ExplorerInfo, 'xVariableName' | 'yVariableName'>;
+
 export function MeasurementExplorer(props: MeasurementExplorerProps) {
   const {
     measurement: { data },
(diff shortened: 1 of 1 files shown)


src/app/components/MeasurementExplorer.tsx
```diff
@@ -23,20 +25,40 @@ export function MeasurementExplorer(props: MeasurementExplorerProps) {
...
+  */
+  function dropdownVariables(axis: 'x' | 'y') {
+    function formatVar(varKey: string) {
+      const { label, units } = variables[varKey];
+      const formatUnit = units ? ` (${units})` : '';
+      const formatVarKey = `${varKey} - `;
+      return formatVarKey + label + formatUnit;
+    }
+    const { variables } = data[info.dataIndex];
+    const oppositeAxis: XOrY = axis === 'x' ? 'yVariableName' : 'xVariableName';
```
wadjih-bencheikh18, line 50: You can remove `XOrY` type
`oppositeAxis` is already defined as `'yVariableName' | 'xVariableName'`

wadjih-bencheikh18 (changes requested):

wadjih-bencheikh18:
Otherwise LGTM!

wadjih-bencheikh18 (approved):

PR #332, opened 2022-11-09
refactor: save only selectedKind

wadjih-bencheikh18:
closes https://github.com/zakodium-oss/analysis-ui-components/issues/269

### DIFF
--- a/src/app-data/data.helpers.ts
+++ b/src/app-data/data.helpers.ts
@@ -35,10 +35,23 @@ export function getFirstMeasurement(
   return null;
 }
 
-export function getCurrentMeasurement(state: AppState) {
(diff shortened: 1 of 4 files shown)


src/app-data/data.helpers.ts
```diff
@@ -35,10 +35,23 @@ export function getFirstMeasurement(
   return null;
 }
 
-export function getCurrentMeasurement(state: AppState) {
-  const { currentMeasurement } = state.view;
-  if (!currentMeasurement) return null;
-  const { kind, id } = currentMeasurement;
+export function getSelectedKind(state: AppState) {
```
targos, line 38: This function returns a measurement, so the name is off.

PR #338, opened 2022-11-10
test: fix dragZone stories

wadjih-bencheikh18:
closes https://github.com/zakodium-oss/analysis-ui-components/issues/308

### DIFF
--- a/src/components/drop-zone/DropZone.tsx
+++ b/src/components/drop-zone/DropZone.tsx
@@ -60,7 +60,6 @@ function DropZoneContent(
     validator: fileValidator,
     onDrop: handleOnDrop,
   });
-
   const getPropsOptions = useMemo(() => {
     if (onClick) {
       return { onClick };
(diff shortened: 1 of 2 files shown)


src/components/drop-zone/DropZone.tsx
```diff
@@ -120,19 +119,23 @@ function DropZoneContent(
...
             </div>
           ) : isContainer ? null : (
-            <p>Drag and drop your files here, or click to select files</p>
+            <p
+              css={css`
+                position: absolute;
+                top: 50%;
+                left: 50%;
+                transform: translate(-50%, -50%);
+              `}
```
stropitek, line 128: Can you put this in a variable since it's a duplication of the css just above

wadjih-bencheikh18: fixed thank you

lpatiny:
@wadjih-bencheikh18 could we close this issue with the same PR ? https://github.com/zakodium-oss/analysis-ui-components/issues/161

PR #343, opened 2022-11-11
fix: log errors from loader

santimirandarp:
This is currently the output (I tweaked one line that is failing in the getMeasurements, already filed the bug.):

* File loading
Screenshot from 2022-11-11 12-31-06

* File errors (unsupported techniques.)
Screenshot from 2022-11-11 12-31-18

@lpatiny 

We may need to reduce the number of files in the link as it is several MBs and xhr requests ? Takes a while here at least.


santimirandarp:
rebased and added in #347

PR #344, opened 2022-11-11
fix: add measurements

wadjih-bencheikh18:
closes https://github.com/zakodium-oss/analysis-ui-components/issues/342

### DIFF
--- a/src/app-data/appState.tsx
+++ b/src/app-data/appState.tsx
@@ -15,6 +15,7 @@ import {
   mergeMeasurements,
   getEmptyDataState,
   MeasurementKind,
+  kindsLabel,
 } from './DataState';
 import {
   getFirstMeasurementOrFail,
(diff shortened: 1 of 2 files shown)


src/app-data/appState.tsx
```diff
@@ -100,22 +101,22 @@ function actionHandler(draft: Draft<AppState>, action: AppStateAction) {
       return;
     case 'ADD_MEASUREMENTS': {
       mergeMeasurements(draft.data.measurements, action.payload);
-      for (let kind of Object.keys(action.payload) as MeasurementKind[]) {
-        if (!draft.view.selectedMeasurements[kind]) {
-          const measurement = getFirstMeasurementOrFail(
+
+      for (const kind of Object.keys(kindsLabel).filter(
+        (k) => k in action.payload,
+      ) as MeasurementKind[]) {
```
targos, line 107: What's the problem with `Object.keys(action.payload)` ?

wadjih-bencheikh18: it's not ordered, in Measurement panel we use the kinds order of kindsLabel

wadjih-bencheikh18: when we add measurements if no one is selected we have to select the first one in the first kind

targos: OK, thank you for the explanation!

PR #349, opened 2022-11-11
refactor: harmonise loaders

santimirandarp:
The parsers were rewritten with this structure (metacode):
```
anyParser(filecollection, logger?:boolean){
 measurements = getEmptyMeasurements()
  logs = [ ] 
  for(loop over files){
   try{
     // try to parse
   } catch(e){
    // push to logs
   }
  return measurements //with the new ones
}
```

The code changes some previous additions from [4ff5580](https://github.com/zakodium-oss/analysis-ui-components/commit/4ff5580a5818a83bd88c078989a7397c45ee577e), just `assert` for an error log and Partial for full `Measurement` object, that's why I included the template, in case it needs discussion.

### DIFF
--- a/src/app-data/DataState.ts
+++ b/src/app-data/DataState.ts
@@ -63,11 +63,12 @@ export const kindsLabel: Record<MeasurementKind, string> = {
 
 export type Loader = (
   fileCollection: FileCollection,
-) => Promise<Partial<Measurements>>;
+  logger?: boolean,
+) => Promise<Measurements>;
 
 export function mergeMeasurements(
-  measurements: Partial<Measurements>,
-  newMeasurements: Partial<Measurements>,
+  measurements: Measurements,
+  newMeasurements: Measurements,
 ) {
   for (const kind in newMeasurements) {
     if (!measurements[kind]) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "@lukeed/uuid": "^2.0.0",
         "@popperjs/core": "^2.11.6",
         "@tanstack/react-table": "^8.5.22",
-        "biologic-converter": "^0.4.0",
+        "biologic-converter": "^0.5.0",
         "cheminfo-types": "^1.4.0",
         "filelist-utils": "^1.2.0",
         "immer": "^9.0.16",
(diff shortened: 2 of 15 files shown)


src/app-data/DataState.ts
```diff
@@ -63,11 +63,12 @@ export const kindsLabel: Record<MeasurementKind, string> = {
...
   fileCollection: FileCollection,
-) => Promise<Partial<Measurements>>;
+  logger?: boolean,
+) => Promise<Measurements>;
 
 export function mergeMeasurements(
-  measurements: Partial<Measurements>,
-  newMeasurements: Partial<Measurements>,
+  measurements: Measurements,
+  newMeasurements: Measurements,
```
targos, line 71: Please don't revert this improvement.

targos:
It looks like you reverted the changes from https://github.com/zakodium-oss/analysis-ui-components/commit/4ff5580a5818a83bd88c078989a7397c45ee577e.

src/app-data/DataState.ts
santimirandarp: If all the loaders return `Measurements` wouldn't it need to be removed? Previously there were some returning Partial Measurements, and some the full object.

santimirandarp:
Not sure whether we could change this

> Error: src/app-data/appState.tsx(103,50): error TS2345: Argument of type 'Partial<Measurements>' is not assignable to parameter of type 'Measurements'.

To accept Measurements, which seems the error. 

> It looks like you reverted the changes from [4ff5580](https://github.com/zakodium-oss/analysis-ui-components/commit/4ff5580a5818a83bd88c078989a7397c45ee577e).

I think some of them yes, that I why I included the template, in case it needs discussion, basically I put all of them to return the same object as that seems easier from my point of view at least.

src/app-data/DataState.ts
targos: All loaders should return `Partial<Measurements>`.

santimirandarp: I see. Basically all those changes were triggered by assuming that `const newMeasurements =  getEmptyMeasurement()` was correct and would help making all the loaders the same, in which case the return would be Measurements. I will try to understand better what it was intended.

PR #354, opened 2022-11-14
refactor: use a template for the loaders

santimirandarp:
* The loaders have now a more similar structure
* Added a few corrections to the default technique in biologic
* Added a `logger?:boolean` key to loaders and load function (maybe this needs to be further up?)

For example, this is how it shows the error for biologic:
Screenshot from 2022-11-14 14-12-33

**Question**

As a user it would be simpler if the measurement explorer labels are in the same side than the axis:

image

### DIFF
--- a/src/app-data/__tests__/getIRMeasurement.ts
+++ b/src/app-data/__tests__/getIRMeasurement.ts
@@ -22,5 +22,5 @@ export async function getIRMeasurement() {
     loaders,
     enhancers,
   });
-  return measurements.ir.entries[0];
+  return measurements?.ir.entries[0];
 }
--- a/src/app-data/__tests__/getIRMeasurement.ts
+++ b/src/app-data/__tests__/getIRMeasurement.ts
@@ -1,5 +1,4 @@
 import { getTestFileCollection } from '../../utils/test-utils';
-import type { IRMeasurement } from '../IRMeasurement';
 import { getIRAutoPeakPickingEnhancer } from '../enhancers/irAutoPeakPickingEnhancer';
 import { irMeasurementEnhancer } from '../enhancers/irMeasurementEnhancer';
 import { loadMeasurements } from '../loadMeasurements';
@@ -23,5 +22,5 @@ export async function getIRMeasurement() {
     loaders,
     enhancers,
   });
-  return measurements.ir?.entries[0] as IRMeasurement;
+  return measurements.ir.entries[0];
 }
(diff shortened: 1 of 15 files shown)


src/app-data/__tests__/getIRMeasurement.ts
```diff
@@ -23,5 +22,5 @@ export async function getIRMeasurement() {
     loaders,
     enhancers,
   });
-  return measurements.ir?.entries[0] as IRMeasurement;
+  return measurements?.ir.entries[0];
```
targos, line 25: I don't think this should be changed.

src/app-data/__tests__/loadMeasurements.test.ts
```diff
@@ -23,7 +23,6 @@ test('loadMeasurements function', async () => {
     enhancers,
   });
 
-  expect(Object.keys(measurements)).toStrictEqual(['nmr', 'ir', 'uv', 'raman']);
```
targos, line 26: I would keep this expect here.

src/app-data/__tests__/getIRMeasurement.ts
santimirandarp: Thanks, I was wondering what was wrong

src/app-data/loaders/biologicLoader.ts
```diff
@@ -1,56 +1,72 @@
-import { v4 } from '@lukeed/uuid';
```
targos, line 1: Why remove the id generation?

santimirandarp: It is done by the template, as most loaders are returning the same data built from the `file`, I can remove it if it makes unclear.

targos: I see. The `templateFromFile` name confused me. Could you rename it to something more descriptive of what it does? Maybe `getMeasurementInfoFromFile()` ?

targos:
> Added a logger?:boolean key to loaders and load function (maybe this needs to be further up?)

Yes, please move to the options of https://github.com/zakodium-oss/analysis-ui-components/blob/f2c23f8952f036e230e2a02706ade65bd60c1cdc/src/app-data/loadMeasurements.ts#L11

src/app-data/loaders/biologicLoader.ts
santimirandarp: Yes. Shall something this simple be tested separately? I thought it would not be necessary but I dont mind to add a `utility/__tests__`

targos: Yes. We should test as much as possible, especially things that seem the most simple, as they usually are simple to test too.

targos:
@santimirandarp Do you think you can finish this PR today? Otherwise I'll take it over because I need it.

santimirandarp:
I'll get back in 2 hours or so, if I cant by then maybe you finish.

PR #395, opened 2022-11-18
feat: add ColorPickerDropdown to the application

Sebastien-Ahkrin:
Close: https://github.com/zakodium-oss/analysis-ui-components/issues/370


src/components/color-picker/gradient-preview/GradiantPreview.tsx
```diff
@@ -0,0 +1,26 @@
+import styled from '@emotion/styled';
```
targos, line 1: Typo in the file name.

targos: Also please move the file to the `gradient-select` folder as it's an internal component.

PR #397, opened 2022-11-19
chore(tsconfig.json): flag usage of implicit any

### LINKED ISSUE #392: Enable noImplicitAny in tsconfig

santimirandarp:
Switch [`noImplicitAny`](https://www.typescriptlang.org/tsconfig#noImplicitAny) value to `true` for flagging `any`s. 

Close #392


package.json
```diff
@@ -101,6 +101,7 @@
     "@ladle/react": "^2.4.5",
     "@playwright/experimental-ct-react": "^1.28.0",
     "@playwright/test": "^1.28.0",
+    "@types/babel__core": "^7.1.20",
```
targos, line 104: Do we really need this dependency? I'm surprised.

santimirandarp: I will double check and put an image of the output just in case I miss something. 

Not sure whether I can modify the rest but I am just trying for learning.

santimirandarp: @targos this is why i had installed it:

```
Could not find a declaration file for module '@babel/core'. '~/analysis-ui-components/node_modules/@babel/core/lib/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/babel__core` if it exists or add a new declaration (.d.ts) file containing `declare module '@babel/core';`

1 import { TransformOptions, ParserOptions } from '@babel/core';
                                                  ~~~~~~~~~~~~~
```

targos: We need to remove `allowJs` to fix that.

targos: I just realized that I was wrong, because the part of the error you printed only included a reference to a `.js` file, but actually the error is triggered by `node_modules/@vitejs/plugin-react/dist/index.d.ts`, which is in TypeScript. So I guess we have no other choice than installing the types for `@babel/core`...

santimirandarp:
Thanks for the explanations. Find them useful. Not in the laptop y&t.

PR #402, opened 2022-11-23
update style of measurements table

Sebastien-Ahkrin:
Close: https://github.com/zakodium-oss/analysis-ui-components/issues/380
Close: https://github.com/zakodium-oss/analysis-ui-components/issues/387
Close: https://github.com/zakodium-oss/analysis-ui-components/issues/390

Capture d’écran 2022-11-23 à 12 06 04

### DIFF
--- a/src/app-data/state/producers/measurements.ts
+++ b/src/app-data/state/producers/measurements.ts
@@ -1,3 +1,4 @@
+import { assertUnreachable } from '../../../components/index';
 import {
   getFirstMeasurementOrFail,
   getMeasurementOrFail,
@@ -43,7 +44,7 @@ export const addMeasurements: AppStateProducer<'ADD_MEASUREMENTS'> = (
   }
 };
 
-export const selectOrUnselectAllMeasurement: AppStateProducer<
+export const selectOrUnselectAllMeasurements: AppStateProducer<
   'SELECT_ALL_MEASUREMENTS'
 > = (draft, action) => {
   const {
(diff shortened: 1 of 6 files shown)


src/app-data/state/producers/measurements.ts
```diff
@@ -43,19 +43,50 @@ export const addMeasurements: AppStateProducer<'ADD_MEASUREMENTS'> = (
   }
 };
 
+export const selectOrUnselectAllMeasurement: AppStateProducer<
```
targos, line 46:
```suggestion
export const selectOrUnselectAllMeasurements: AppStateProducer<
```

src/app-data/state/producers/measurements.ts
```diff
@@ -43,19 +43,50 @@ export const addMeasurements: AppStateProducer<'ADD_MEASUREMENTS'> = (
...
+        (element) => element !== id,
+      );
+      break;
+    }
+    case 'replace': {
+      draft.view.selectedMeasurements[kind] = [id];
+      break;
+    }
+    default:
+      break;
```
targos, line 88:
```suggestion
      assertUnreachable(acc);
```

src/app/panels/measurements/MeasurementsTable.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+export function MeasurementsTable(props: MeasurementsTableProps) {
+  const { kind } = props;
+
+  const {
+    data: { measurements },
+  } = useAppState();
+
+  return (
+    <table css={measurementsTableCss.root}>
```
targos, line 103: We will merge this for now, but in the future we will have to use the Table component again, so the improvements need to be in the Table component, not only in MeasurementsTable.

PR #404, opened 2022-11-24
fix: resolve error on close items

Sebastien-Ahkrin:
close: https://github.com/zakodium-oss/analysis-ui-components/issues/372

### DIFF
--- a/src/components/dropdown-menu/MenuItems.tsx
+++ b/src/components/dropdown-menu/MenuItems.tsx
@@ -126,7 +126,11 @@ function Item<T>(props: ItemProps<T>) {
     <Menu.Item disabled={option.disabled}>
       {({ active }) => (
         <ItemDiv
-          onClick={() => onSelect(option)}
+          onClick={() => {
+            if (!option.disabled) {
+              onSelect(option);
+            }
+          }}
           active={active}
           disabled={option.disabled || false}
         >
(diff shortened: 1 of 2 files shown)


targos:
Please think about your commit messages: they should explain what you do. Here there was no error. I reworded to `fix: do not close DropdownMenu when disabled items are clicked`

targos:
The problem now is that it also doesn't close when you click on enabled items...

Sebastien-Ahkrin:
> The problem now is that it also doesn't close when you click on enabled items...

Same as Dropdown with context.
In fact, the thing that make the menu close was a bug (with the clickOutside), so do we have to call the close function on both of the components ?

targos:
What is Dropdown with context?

Sebastien-Ahkrin:
> What is Dropdown with context?

the one with the right click

targos:
Yes, we have to call the close function in both.

src/components/dropdown-menu/MenuItems.tsx
```diff
@@ -126,7 +126,11 @@ function Item<T>(props: ItemProps<T>) {
     <Menu.Item disabled={option.disabled}>
       {({ active }) => (
         <ItemDiv
-          onClick={() => onSelect(option)}
+          onClick={() => {
+            if (!option.disabled) {
+              onSelect(option);
+            }
+          }}
```
targos, line 133: should we remove this now that you call it above?

targos: I mean remove everything. You already call `onSelect` above.

targos: @Sebastien-Ahkrin

Sebastien-Ahkrin: i'm never calling onSelect if i remove theses line

Issue #326, opened 2022-11-08
Refactor loaders to be kind-based instead of extension-based


targos:
In the future, it will be a nightmare to configure the application if you need to know which loader is capable of loading which kind of measurement. Then if the application is supposed to only load IR spectra, we would have to add options to all loaders so they ignore other kinds of data.

PR #415, opened 2022-11-27
Load parsers by kind

santimirandarp:
This is an idea related to #326, still a draft.

* `loadByKind` maps `MeasurementKind`s to an array of loaders

Then a user wouldn't need to know name of the loaders to use it.


targos:
Please don't go too far with this. It's something we wanted to discuss before writing any code.

santimirandarp:
> Please don't go too far with this. It's something we wanted to discuss before writing any code.

No problem, I can drop the PR if this is not the right time or code.

PR #419, opened 2022-11-28
feat: allow to change all measurements view at once

Sebastien-Ahkrin:
close: https://github.com/zakodium-oss/analysis-ui-components/issues/391

### DIFF
--- a/src/app-data/state/producers/measurements.ts
+++ b/src/app-data/state/producers/measurements.ts
@@ -124,17 +124,17 @@ export const selectMeasurementKind: AppStateProducer<
 export const setMeasurementVisibility: AppStateProducer<
   'SET_MEASUREMENT_VISIBILITY'
 > = (draft, action) => {
-  if (action.payload.id === null) {
-    for (const id of Object.keys(draft.view.measurements)) {
-      const measurementView = draft.view.measurements[id];
-      measurementView.visible = action.payload.isVisible;
-    }
-
-    return;
-  }
-
   const measurementView = draft.view.measurements[action.payload.id];
   measurementView.visible = action.payload.isVisible;
+};
+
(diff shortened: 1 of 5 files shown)


src/app-data/state/producers/measurements.ts
```diff
@@ -108,6 +108,15 @@ export const selectMeasurementKind: AppStateProducer<
 export const setMeasurementVisibility: AppStateProducer<
   'SET_MEASUREMENT_VISIBILITY'
 > = (draft, action) => {
+  if (action.payload.id === null) {
+    for (const id of Object.keys(draft.view.measurements)) {
```
targos, line 112: You should only update the selected kind, like with "select all"

Sebastien-Ahkrin: do i have to update, only the selected items ?

targos: yes

Sebastien-Ahkrin: okay, so i will mark this conversation as resolved :)

targos: Actually, we discussed last week and the issue was wrong (I just updated it)
There should be two separate buttons with the eyes (to show and hide). Click on the buttons should only apply to the selected measurements.

For example, with this selection:
image

After the click to the open eye, it should be like this:
image

src/app-data/state/producers/measurements.ts
```diff
@@ -129,7 +129,7 @@ export const setMeasurementVisibility: AppStateProducer<
 };
 
 export const setAllMeasurementVisibility: AppStateProducer<
```
targos, line 131: rename this too

targos:
Layout is broken: 
image

PR #429, opened 2022-11-30
feat add input

Sebastien-Ahkrin:
Parts of: https://github.com/zakodium-oss/analysis-ui-components/issues/346

- [x] Controlled component
- [x] Forwards the native input props
- [x] Placeholder
- [x] 2 sizes
- [x] With / without labels
- [x] Size on Field too
- [x] trailing / leading component (inline or not)
- [ ] validation feedback
- [x] handle required props

### DIFF
--- a/src/components/forms/Input.tsx
+++ b/src/components/forms/Input.tsx
@@ -5,11 +5,8 @@ import { useFieldsContext } from './context/FieldsContext';
 
 interface StyledProps {
   variant: 'default' | 'small';
-  hasTrailing: boolean;
   hasLeading: boolean;
-
-  hasInlineTrailing: boolean;
-  hasInlineLeading: boolean;
+  hasTrailing: boolean;
 }
 
 function getSpecialSize(props: Pick<StyledProps, 'variant'>) {
@@ -19,11 +16,7 @@ function getSpecialSize(props: Pick<StyledProps, 'variant'>) {
   };
 }
 
-const LabelStyled = styled.label<{
-  hasLeading: boolean;
-  hasTrailing: boolean;
-  variant: 'small' | 'default';
-}>`
+const LabelStyled = styled.label<StyledProps>`
   padding: ${(props) =>
     props.variant === 'default'
       ? props.hasTrailing
--- a/src/components/forms/Input.tsx
+++ b/src/components/forms/Input.tsx
@@ -117,23 +117,18 @@ export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
 }
 
 export function Input(props: InputProps) {
-  const {
-    variant = 'default',
-    trailingAddon,
-    leadingAddon,
-    ...otherProps
-  } = props;
+  const { variant, trailingAddon, leadingAddon, ...otherProps } = props;
 
   const { name, variant: contextVariant } = useFieldsContext();
 
   return (
     <GroupStyled>
-      {renderLeadingAddon(variant, leadingAddon)}
+      {renderLeadingAddon(variant || contextVariant, leadingAddon)}
 
       <InputStyled
         id={name}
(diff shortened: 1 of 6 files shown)


stories/components/input.stories.tsx
```diff
@@ -0,0 +1,39 @@
+import { ChangeEvent, useState } from 'react';
+
+import { Input } from '../../src/components/forms/Input';
+import {
+  Field,
+  Fields,
+} from '../../src/components/forms/context/FieldsContext';
```
targos, line 7: all should be imported from `../../src/components/index`

stropitek:
You can add the `required` prop to the `Field` component? There is an example in the issue of how a required field can look like.

stropitek:
- Story with small and regular sizes next to each other
- Story with small and regular sizes next to each other but with label (should work by setting a `size` prop to `Field` not `Input`

src/components/forms/Input.tsx
```diff
@@ -0,0 +1,40 @@
...
+  min-width: 0px;
+  background-color: #fff;
+  border-radius: ${(props) => (props.inputSize === 'default' ? '6px' : '4px')};
+  transition: all 0.2s;
+  border: 1px solid #d9d9d9;
+`;
+
+export interface InputProps
+  extends Omit<InputHTMLAttributes<HTMLInputElement>, 'size'> {
+  inputSize?: 'default' | 'small';
```
stropitek, line 25: I neither like highjacking the `size` prop nor the `inputSize` name. Maybe call it variant?

Sebastien-Ahkrin:
Do i have to add a "size" props on the FieldContext for Input @stropitek  ?

stropitek:
> Do i have to add a "size" props on the FieldContext for Input

yep. If size set directly on input it takes precedence

stropitek:
@Sebastien-Ahkrin what is the state of this PR?

Sebastien-Ahkrin:
> @Sebastien-Ahkrin what is the state of this PR?

@stropitek i'm working on Leading & Trailing addon actually, but actually i can make the pull request okay to check & merge. And i finalize my work on Trailing on a new pull request

stropitek:
Did you already work on the latest feedback that was given (no commits after them)?



> i'm working on Leading & Trailing addon actually

Yes please don't include that in this PR

stropitek:
image

On Edge (the browser) there is the default black outline. It should be removed.

stropitek:
If the size prop is not used to change the appearance of the label, then there is no need to have it on `Field`

stropitek:
The outline on the input should stay when the input is focused

stropitek:
The label should have the same font size as the input.

Therefore I think it makes sense to add the `variant` prop to the `Field` component. It can be inherited by the input via the context.

src/components/forms/Input.tsx
```diff
@@ -133,7 +133,7 @@ export function Input(props: InputProps) {
       <InputStyled
         id={name}
         name={name}
-        variant={variant}
+        variant={contextVariant || variant}
```
stropitek, line 136: It should be `variant` || `contextVariant` and `variant` should not have a default value, but `contextVariant` should

stropitek:
image

Sebastien-Ahkrin:
> image

@stropitek what navigator are you using and what's the name of the stories for this screen ?

src/components/forms/Input.tsx
```diff
@@ -24,7 +24,13 @@ const GroupStyled = styled.div`
   }
 `;
 
-const lineHeight = 4 / 7;
+function getSpecialSize(props: any) {
```
stropitek, line 27: 👀

stropitek (approved):

PR #442, opened 2023-01-05
refactor: change style and css to use styled component

Sebastien-Ahkrin:


src/components/button/ButtonGroup.tsx
```diff
@@ -60,12 +46,12 @@ ButtonGroup.Button = function ButtonGroupButton(props: ButtonGroupButtonProps) {
   const { position, label, onClick } = props;
 
   return (
-    <button
+    <ButtonGroupCustomButton
       type="button"
       onClick={onClick}
-      css={position === 'first' ? styles.firstButton : styles.secondButton}
+      second={position === 'last'}
```
stropitek, line 52: Why call it "second" and not "last" or "isLast" when it's the last?

stropitek (approved):

Issue #445, opened 2023-01-19
Stop click propagation on MenuItems


hamed-musallam:
@targos 

in NMRium, the spectra table column does two functionality 
1- order rows (Asc,Dec)
2- open the context menu

the click event on MenuItem trigger accidentally the order action, can we stop propagating the click event on the MenuItem ?

https://github.com/zakodium-oss/react-science/blob/0637f05eecf0c48c23a1ded7a78e7badeb00d35c/src/components/dropdown-menu/MenuItems.tsx#L129-L131

targos:
yes

PR #446, opened 2023-01-19
fix: stop click event propagation on MenuItem

hamed-musallam:
closes: https://github.com/zakodium-oss/react-science/issues/445


src/components/dropdown-menu/MenuItems.tsx
```diff
@@ -126,7 +126,8 @@ function Item<T>(props: ItemProps<T>) {
     <Menu.Item disabled={option.disabled}>
       {({ active }) => (
         <ItemDiv
-          onClick={() => {
+          onClick={(e) => {
+            e.stopPropagation();
```
targos, line 130:
```suggestion
          onClick={(event) => {
            event.stopPropagation();
```

Issue #452, opened 2023-03-03
Splitter needs to observer the changes in the initialSize and initialClosed properties.


hamed-musallam:
We need the SplitterPane component to re-render when the `initialSize` and `initalClosed` properties changed

hamed-musallam:
can we use `useEffect` to listen for initial value changes and update the initialSize and initalClosed values

https://github.com/zakodium-oss/react-science/blob/main/src/components/hooks/useOnOff.ts


```ts
useEffect(()=>{
  setOnOff(initialValue);
),{initialValue})
```

https://github.com/zakodium-oss/react-science/blob/993f5a1f615fb90ac3241810f7aa812a4414b313/src/components/split-pane/SplitPane.tsx#L99


```ts
useEffect(()=>{
 const size = parseSize(initialSize);
  setSize(size);
),{initialSize})
```

hamed-musallam:
Currently, the only ways to update these values after it is initialized is by forcing the Splitter to re-render by changing the `key` property or refreshing the page

hamed-musallam:
we have two options to keep these properties `initialSize` and `initalClosed` as it is and rename them to `size` and `closed` or we could keep them and create new properties and by that, we have `initialSize` and `initalClosed` when we want just an initial value and `size` and `closed` which will track all the changes

PR #457, opened 2023-03-07
fix: splitter not observe 'size' and 'closed' values changes

hamed-musallam:
- [X] https://github.com/zakodium-oss/react-science/issues/452

### DIFF
--- a/src/components/split-pane/SplitPane.tsx
+++ b/src/components/split-pane/SplitPane.tsx
@@ -10,6 +10,7 @@ import {
   useState,
   RefObject,
   useReducer,
+  useEffect,
 } from 'react';
 import useResizeObserver from 'use-resize-observer';
 
@@ -41,6 +42,7 @@ export interface SplitPaneProps {
    * @default '50%'
    */
   initialSize?: SplitPaneSize;
+  size?: SplitPaneSize;
   /**
    * Defines whether the pane is initially closed.
    * A value of `true` means the pane is always initially closed.
@@ -53,6 +55,7 @@ export interface SplitPaneProps {
    * @default false
    */
   initialClosed?: boolean | number;
+  closed?: boolean | number;
   /**
    * Called whenever the user finishes resizing the pane.
    */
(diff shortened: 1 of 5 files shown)


stropitek:
> /cc @stropitek given you recently had a splitter library on your radar.

The library I was looking at is:
https://github.com/bvaughn/react-resizable-panels

I looked at it and one of the main limitation is that all sizes are expressed in percent. It does not seem possible to set a default panel size in pixel units, which I think we do in NMRium.

targos:
https://github.com/bvaughn/react-resizable-panels/issues/47#issuecomment-1368108429

Please also review!

stropitek:
> Please also review!

I don't think I'm ok with the API changes. I don't understand the point of having both `size` / `initialSize`, `closed` / `initialClosed` for properties which are uncontrolled.

The issue also is not very helpful because it describes the solution and not the problem.

@lpatiny will show me the root cause of the bug and I will make a second assessment.

@hamed-musallam if you could explain the original problem in the issue that would be helpful.

hamed-musallam:
Currently, in NMRium we can control the size of the split panel from the general settings and it is per workspace and we always update the value once the `resize` is ends, the problem came when we switch between workspaces or change from the general setting the Splitter will not recognize the new value and only way is by forcing the Splitter to re-render by changes the `key` value, The same applied to the `closed` prosperity.

if we always change the `key` value this lead to re-render all the children.


What we have right now in this PR, is we keep default properties `initialClosed` and `intialSize`, and if we want to control these values from the parent component we can use `size` and `closed` properties

stropitek:
> Take a look at this comment [#452 (comment)](https://github.com/zakodium-oss/react-science/issues/452#issuecomment-1457919090)

Why did you choose to add properties rather than keep (or rename) the `initialSize` and `initialClosed` properties?

hamed-musallam:
Just to keep the old component interface as it is because most of the users will not control the `closed` and `size` properties which I think is a special case in NMRium so they will hardcode the values for these properties and will not change it.

hamed-musallam:
I removed the  `initialSize` and `initialClosed` properties and keep the  `size` and `closed` properties and made the required changes, can you check the changes and if everything from your side looks good to you, can you merge it and create a new release from the package

stropitek (approved):
LGTM but I can't test the storybook. Next time please make sure to create a branch on the original repo rather than on a fork so that the cloudflare build works.

Issue #475, opened 2023-04-25
Context menu not triggering if we add padding on the table column or if the column is empty


hamed-musallam:

stropitek:
Is the component with padding above or below the context menu in the component tree?

hamed-musallam:
the component with padding is above the context menu 

_**td (padding) > DropdownMenu> text**_ 

 the DropdownMenu is on each cell in the row of the table to have its own data for each row.

Is there any way to have DropdownMenu over the table that listens for only the rows context menu event instead of having it per Cell?

hamed-musallam:
> Move the padding below DropdownMenu. In principle possible but might not be practical in your case. I think it works to create divs with padding within a td.

I think this way is not practical and we need to look into the second option that you proposed to adapt the DropdownMenu, Maybe we can have 
 ```
 < DropdownMenuContext >
   <table />
</ DropdownMenuContext >   

```
and pass down the onContext call-back to dom element that we need to have context on it

targos:
Why do you want to wrap the whole table (or individual cells like it seems to be the case now) instead of rows if your intent is to have row-based behavior? That doesn't make sense to me.

stropitek:
> Why do you want to wrap the whole table (or individual cells like it seems to be the case now) instead of rows if your intent is to have row-based behavior? That doesn't make sense to me.

We clarified this in a zoom call. @hamed-musallam is going to work on a PR with the `as` props as suggested in https://github.com/zakodium-oss/react-science/issues/475#issuecomment-1522918804

hamed-musallam:
it seems that I was wrong, The Daniel suggestion is better,  we can have an option to render the DropdownMenu as `tr` or `td`, the default one will be div, and this will solve our problem.

hamed-musallam:
closed by https://github.com/zakodium-oss/react-science/pull/476

PR #476, opened 2023-04-26
feat: adapt the DropdownMenu to accept 'as' prop to specify an element to be a wrapper

hamed-musallam:
- [X] https://github.com/zakodium-oss/react-science/issues/475

### DIFF
--- a/src/components/dropdown-menu/DropdownMenu.tsx
+++ b/src/components/dropdown-menu/DropdownMenu.tsx
@@ -1,6 +1,6 @@
 import { Menu } from '@headlessui/react';
 import type { Placement } from '@popperjs/core';
-import { ReactNode, useRef } from 'react';
+import { ReactNode, useRef, ElementType, ComponentProps } from 'react';
 
 import { useModifiedPopper } from '../hooks/useModifiedPopper';
 import { useOnClickOutside } from '../hooks/useOnClickOutside';
@@ -22,8 +22,8 @@ interface DropdownMenuBaseProps<T> {
 
 type ElementTags = keyof JSX.IntrinsicElements;
 
-type ElementProps<Tag> = Tag extends ElementTags
-  ? JSX.IntrinsicElements[Tag]
+type ElementProps<T = unknown> = T extends ElementType
+  ? ComponentProps<T>
   : never;
 
 interface DropdownMenuClickProps<T> extends DropdownMenuBaseProps<T> {
@@ -45,7 +45,7 @@ export type DropdownMenuProps<T> =
   | DropdownMenuContextProps<T>
   | DropdownMenuClickProps<T>;
 
-export function DropdownMenu<T = unknown, E extends ElementTags = 'div'>(
+export function DropdownMenu<T = unknown, E extends ElementType = 'div'>(
   props: DropdownMenuProps<T> & { wrapperProps?: ElementProps<E> },
 ) {
   const { trigger, ...otherProps } = props;
@@ -58,7 +58,7 @@ export function DropdownMenu<T = unknown, E extends ElementTags = 'div'>(
   );
 }
 
-function DropdownContextMenu<T, E extends ElementTags = 'div'>(
+function DropdownContextMenu<T, E extends ElementType = 'div'>(
   props: Omit<DropdownMenuContextProps<T>, 'trigger'> & {
     wrapperProps?: ElementProps<E>;
   },
@@ -83,7 +83,7 @@ function DropdownContextMenu<T, E extends ElementTags = 'div'>(
   const ref = useRef<HTMLDivElement>(null);
   useOnClickOutside(ref, closePopperElement);
 
-  const { style = {}, ...otherWrapperProps } = wrapperProps || { style: {} };
+  const { style = {}, ...otherWrapperProps } = wrapperProps as ElementProps<E>;
   return (
     <Wrapper
       style={{ ...(props?.as === 'div' && { display: 'contents' }), ...style }}
--- a/src/components/dropdown-menu/DropdownMenu.tsx
+++ b/src/components/dropdown-menu/DropdownMenu.tsx
@@ -20,8 +20,8 @@ interface DropdownMenuBaseProps<T> {
   onSelect: (selected: MenuOption<T>) => void;
 }
 
-type ElementProps<T = unknown> = T extends ElementType
-  ? ComponentProps<T>
+type ElementProps<E = unknown> = E extends ElementType
+  ? ComponentProps<E>
   : never;
 
 interface DropdownMenuClickProps<T> extends DropdownMenuBaseProps<T> {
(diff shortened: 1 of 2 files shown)


hamed-musallam:
How we can only have the HTML Element Props ?   
https://github.com/zakodium-oss/react-science/blob/a2729e6c6cd722af5ebd723dfffbdd95236b9298/src/components/dropdown-menu/DropdownMenu.tsx#L23-L27

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -32,30 +37,39 @@ interface DropdownMenuClickProps<T> extends DropdownMenuBaseProps<T> {
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  as?: ElementTags | React.ComponentType<any>;
 }
 
 export type DropdownMenuProps<T> =
   | DropdownMenuContextProps<T>
   | DropdownMenuClickProps<T>;
 
-export function DropdownMenu<T>(props: DropdownMenuProps<T>) {
+export function DropdownMenu<T = unknown, E extends ElementTags = 'div'>(
```
targos, line 48: `E extends ElementType` with `import { ElementType } from 'react';

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -21,6 +20,12 @@ interface DropdownMenuBaseProps<T> {
   onSelect: (selected: MenuOption<T>) => void;
 }
 
+type ElementTags = keyof JSX.IntrinsicElements;
+
+type ElementProps<Tag> = Tag extends ElementTags
+  ? JSX.IntrinsicElements[Tag]
+  : never;
```
targos, line 27:
```
type ElementProps<T = unknown> = T extends ElementType
  ? ComponentProps<T>
  : never;
```

With `import { ElementType, ComponentProps } from 'react';`

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -32,30 +37,39 @@ interface DropdownMenuClickProps<T> extends DropdownMenuBaseProps<T> {
...
-const HandleMenuContextDiv = styled.div`
-  display: contents;
-`;
-
-function DropdownContextMenu<T>(props: Omit<DropdownMenuProps<T>, 'trigger'>) {
-  const { children, onSelect, ...otherProps } = props;
+function DropdownContextMenu<T, E extends ElementType = 'div'>(
+  props: Omit<DropdownMenuContextProps<T>, 'trigger'> & {
+    wrapperProps?: ElementProps<E>;
+  },
```
stropitek, line 64: Can you flatten the prop structure by putting the props from the `as` element at the first level? I don't think we will ever run into name clashes.

```suggestion
  props: Omit<DropdownMenuContextProps<T>, 'trigger'> & ElementProps<E>
```

hamed-musallam: i did it like that because we have props for the `Menu` and others for `Wrapper` unless we destruct all the menu pros at the top level 
https://github.com/zakodium-oss/react-science/blob/5286bc0db1d645a94733a41574c0a94ba9543049/src/components/dropdown-menu/DropdownMenu.tsx#L109

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -69,33 +83,40 @@ function DropdownContextMenu<T>(props: Omit<DropdownMenuProps<T>, 'trigger'>) {
...
-                  itemsStatic
-                  onSelect={(selected) => {
-                    closePopperElement();
-                    onSelect(selected);
-                  }}
-                  {...otherProps}
-                />
-              </Menu>
+    <Wrapper
+      style={{ ...(props?.as === 'div' && { display: 'contents' }), ...style }}
```
stropitek, line 89: I think you are breaking the behavior when `as` is undefined

It's up to the user to specify everything including the style if they chose to pass the `as` prop.

stropitek: Sorry it is not breaking because of the default value.

But I think when the user sets the as prop to `div`, the component should not assume the user want `display: 'contents'` on it

hamed-musallam: if the user does not pass the `as` props, I assume he needs to have the old behavior

src/components/dropdown-menu/DropdownMenu.tsx
hamed-musallam: We can destructuring the `Menu` props but we have to take care of any coming changes on the component has to take care about this (I mean for the menu props)

src/components/dropdown-menu/DropdownMenu.tsx
targos: I agree with Daniel. We should only set `display: 'contents'` if `props.as` is not defined.

hamed-musallam: > Sorry it is not breaking because of the default value.
> 
> But I think when the user sets the as prop to `div`, the component should not assume the user want `display: 'contents'` on it

it depends on how we want the component to behave

stropitek: It's a question giving meaning and coherence to the props.
not specifying `as` = "I let the component decide the what to render in a way that generally works"
specifying `as` = "I take over control and I decide exactly what the wrapper element should be and the value of its props, including the style attribute"

hamed-musallam:
Can you merge this PR and we can have a new release Today?

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -32,30 +35,40 @@ interface DropdownMenuClickProps<T> extends DropdownMenuBaseProps<T> {
 interface DropdownMenuContextProps<T> extends DropdownMenuBaseProps<T> {
   trigger: 'contextMenu';
   children: ReactNode;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  as?: ElementType | React.ComponentType<any>;
```
stropitek, line 39: You don't need the `React.ComponentType`, and `as` should be using the `E` generic.

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -32,20 +32,20 @@ interface DropdownMenuClickProps<T> extends DropdownMenuBaseProps<T> {
   trigger: 'click';
 }
 
-interface DropdownMenuContextProps<T> extends DropdownMenuBaseProps<T> {
+interface DropdownMenuContextProps<T, E> extends DropdownMenuBaseProps<T> {
   trigger: 'contextMenu';
   children: ReactNode;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  as?: ElementType | React.ComponentType<any>;
+  as?: E | React.ComponentType<any>;
```
stropitek, line 39: Why keep `React.ComponentType`?

hamed-musallam: it is a mistake, I Just replace the name.

hamed-musallam:
can we have a patch release as soon as possible

PR #484, opened 2023-05-12
Storybook-migration

### LINKED ISSUE #470: Switch from ladle to storybook
I think if you run `npx storybook init` like [documented here](https://storybook.js.org/docs/react/get-started/install) it should detect a vite project and install the necessary dependencies. Not sure what will happen with the existing stories.

Resources that can help you:
- https://storybook.js.org/docs/react/builders/vite#setup
- https://github.com/zakodium/components/blob/main/.storybook/main.js

Make sure that:
- The `build-stories` script puts the build in the same directory as before so that the `build-all` still works (check that the PR preview works)
- You remove all ladle-related dependencies
- You can still control stories which are using `args` / `argTypes` after the migration to storybook. You can use the essentials addon for that: https://storybook.js.org/docs/react/essentials/introduction

moonayyur:
closes #470


package.json
```diff
@@ -26,15 +26,13 @@
   ],
   "scripts": {
     "build": "npm run clean && npm run build-ts && npm run fix-react-shadow",
-    "build-all": "npm run build-app && npm run build-stories && mv build dist/stories",
+    "build-all": "npm run build-app && npm run build-storybook && mv build dist/stories",
```
stropitek, line 29: You can test that this script works locally

```suggestion
    "build-all": "npm run build-app && npm run build-storybook && mv storybook-static dist/stories",
```

moonayyur: It works thanks

moonayyur:
@stropitek The migration is done and stories with controls are still functional.
There are few issues that have emerged :
- The style isn't the same as before for buttons, borders and other elements
- The `onRequestClose` actions aren't functioning properly in **Modal** stories as they don't close the modal
- The `border` control on Table **Row Border** and **Header Border** isn't working as expected, always resulting in a "false" value being passed to the **Table** component

.storybook/main.ts
```diff
@@ -0,0 +1,20 @@
+import type { StorybookConfig } from '@storybook/react-vite';
+const config: StorybookConfig = {
+  stories: ['../stories/**/*.stories.@(js|jsx|ts|tsx)'],
+  core: {},
+  addons: [
+    '@storybook/addon-links',
+    '@storybook/addon-essentials',
+    '@storybook/addon-interactions',
```
stropitek, line 8: You can remove + uninstall this addon, I believe we don't use it

.storybook/main.ts
```diff
@@ -0,0 +1,20 @@
...
+  ],
+  framework: {
+    name: '@storybook/react-vite',
+    options: {
+      strictMode: false,
```
stropitek, line 13: Can you enable the `fastRefresh` option? (I'm not sure if enabled by default)

.storybook/preview.ts
```diff
@@ -0,0 +1,15 @@
+import type { Preview } from '@storybook/react';
+
+const preview: Preview = {
+  parameters: {
+    actions: { argTypesRegex: '^on[A-Z].*' },
```
stropitek, line 5: Interesting!

stories/app/about-dialog.stories.tsx
```diff
@@ -2,9 +2,11 @@ import { FaMeteor } from 'react-icons/fa';
 
 import { AboutDialogToolbarButton } from '../../src/app/index';
 import { Toolbar } from '../../src/components/index';
+import { RootLayoutDecorator } from '../utils';
 
 export default {
   title: 'Components / AboutDialogToolbarButton',
+  decorators: [RootLayoutDecorator],
```
stropitek, line 9: The `RootLayoutDecorator` is required for all stories. Can you set it up to be a global decorator so that you don't have to pass it to every story?

https://storybook.js.org/docs/react/writing-stories/decorators#global-decorators

stropitek:
> The style isn't the same as before for buttons, borders and other elements

This must be due to the absence of the root layout in most stories. The root layout adds CSS reset stylesheets.

stropitek:
> * The `onRequestClose` actions aren't functioning properly in **Modal** stories as they don't close the modal

Could it be because of the rule you setup in `preview.js`, with the regex for argTypes starting with `on`?

stropitek:
I pinned the version of ml-spectra-processing because using typescript's `node16` introduces an issue related to `@headlessui/react`.

We plan to remove this dependency so I suggest we use `node16` after the library has been removed.

package.json
```diff
@@ -98,9 +98,13 @@
...
     "@babel/core": "^7.21.4",
     "@babel/eslint-parser": "^7.21.3",
-    "@ladle/react": "^2.12.3",
     "@playwright/experimental-ct-react": "^1.33.0",
     "@playwright/test": "^1.33.0",
+    "@storybook/addon-essentials": "^7.0.10",
+    "@storybook/blocks": "^7.0.10",
+    "@storybook/react": "^7.0.10",
+    "@storybook/react-vite": "^7.0.10",
+    "@storybook/testing-library": "^0.1.0",
```
targos, line 107: We don't use Testing Library

package.json
```diff
@@ -113,11 +117,14 @@
     "cross-env": "^7.0.3",
     "eslint": "^8.39.0",
     "eslint-config-zakodium": "^7.2.0",
+    "eslint-plugin-storybook": "^0.6.12",
     "prettier": "^2.8.8",
+    "prop-types": "^15.8.1",
```
targos, line 122: Do we use `prop-types` anywhere?

PR #485, opened 2023-05-23
fix(tabs): ensure consistant behavior between horizontal and vertical

Xstoudi:
Without the change, vertical tabs allow their child to work in 100% of the width:
image

but horizontal tabs are limited to their own height:
image

### DIFF
--- a/src/components/tabs/Tabs.tsx
+++ b/src/components/tabs/Tabs.tsx
@@ -152,7 +152,7 @@ function TabsHorizontal<T extends string = string>(
   const item = useTabsContext();
   return (
     <div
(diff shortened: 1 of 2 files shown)


targos:
Can you write a story that demonstrates the fix (or modify an existing one)?

stropitek:
> The failing test is not failing on my computer. Is it flacky?

Yes this test is and you can ignore the failing test on webkit, ct.

I added you to a team which gives you write rights on this repo. You should receive an email invite to join the team. Can you repush this branch on the zakodium repository and replace this PR with a new one so that we have the preview working?

Issue #473, opened 2023-04-05
Create checkbox component


stropitek:
## Features
- Controlled component
- indeterminate state
- with label
- disabled state

## API

- Label can be anything (`ReactNode`)

```tsx
<>
  <Checkbox checked label="Checkbox checked" onChange={(checked: boolean) => {}} />
  <Checkbox checked={false} label="Checkbox unchecked" />
  <Checkbox checked label="Checkbox checked" disabled />
  // If indeterminate is passed doesn't matter if checked is true or not
  <Checkbox checked indeterminate label="Checkbox indeterminate" />
  <Checkbox checked disabled label="Checkbox disabled" />
</>
```

## Design

Compact. The ant design looks nice.
https://ant.design/components/checkbox

![CleanShot 2023-04-05 at 11 01 42](https://user-images.githubusercontent.com/4118690/230034327-e6feb2c9-cc6e-484d-b618-fc3c2541046b.png)

![CleanShot 2023-04-05 at 11 04 41](https://user-images.githubusercontent.com/4118690/230034806-c6bcf59c-eef0-4908-96cf-eb60c30feb94.png)

## Requirements
Use radix-ui checkbox component for accessibility: https://www.radix-ui.com/docs/primitives/components/checkbox#checkbox

PR #490, opened 2023-05-30
feat: checkbox component

Xstoudi:
Notable change compared to issue requirement is that third state is not another boolean value, instead it's a third possible `checked` value, Radix-style. This solution avoid confusion of behavior in case `checked` and `indeterminate` are both `true`.
Tell me if you really prefer to have a separate `indeterminate` value.

Fixes: https://github.com/zakodium-oss/react-science/issues/473

### DIFF
--- a/stories/components/checkbox.stories.tsx
+++ b/stories/components/checkbox.stories.tsx
@@ -1,5 +1,4 @@
 import { useCallback, useMemo, useState } from 'react';
-import { setupChaiConfig } from 'vitest';
 
 import { CheckedState, Checkbox, Input } from '../../src/components/index';
 


stropitek:
> * Add a disabled input in the controlled story

You might actually need to create another story (because otherwise won't be able to test all the possibilities in the existing one)

stropitek:
Noticed something else when doing another pass:
- Can you move the story to the "forms" section
- Can you make another story where you put an input and a checkbox together in a simple vertical layout. Want to see if the font sizes match (with one of the possible size variants)

stropitek:
> Another concern, text in `Input` doesn't seem to be vertically centered. Is it on purpose?

No, I will create an issue for that

Issue #472, opened 2023-04-05
Create Select component


stropitek:
## Features
- Controlled component
- Separators (horizontal lines between lists of options, see [radix-ui example](https://www.radix-ui.com/docs/primitives/components/select#select))
- Categories (label to group options together), see radix-ui and ant examples.
- disabled options
- Keyboard navigation (handled by radix-ui)

## API
- Label can be anything (`ReactNode`)
- Find selected element by comparing the `value` property

```jsx
<Select
  selected={{label: 'Apple, value: 'apple'}}
  onSelect={(option) => {
    console.log(option); // {label: 'Banana', value: 'banana'}
  }

  // First level of array controls where dividers are placed
  options={[
    [
      // options with categories
      {
        label: 'Fruit',
        options: [
          { label: 'Banana', value: 'banana' },
          { label: 'Apple', value: 'apple' },
        ],
      },
      {
        label: 'Vegetables',
        options: [
          { label: 'Carrot', value: 'carrot' },
          { label: 'Broccoli', value: 'broccoli' },
        ]
      }
    ],
    [
      // options without categories
      {label: 'Beef', value: 'beef'},
      {label: 'Pork', value: 'pork', disabled: true},
      {label: 'Chicken', value: 'chicken'},
    ]
  ]}
/>
```

## Design
As for all react-science components, should be designed with compactness in mind. Use blue / gray as the base colors.
You can get inspired by:
- Radix-ui: https://www.radix-ui.com/docs/primitives/components/select#select
- Ant design: https://ant.design/components/select

Requirements:
- Use radix-ui: https://www.radix-ui.com/docs/primitives/components/select#select

stropitek:
Closed by https://github.com/zakodium-oss/react-science/pull/495

PR #495, opened 2023-06-06
feat: add select component

Xstoudi:
Work in progress, I still need to figure out why keyboard navigation is not working.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "@lukeed/uuid": "^2.0.1",
         "@popperjs/core": "^2.11.7",
         "@radix-ui/react-checkbox": "^1.0.4",
+        "@radix-ui/react-select": "^1.2.2",
         "@tanstack/react-query": "^4.29.5",
         "@tanstack/react-table": "^8.8.5",
         "biologic-converter": "^0.6.0",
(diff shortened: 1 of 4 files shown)


stropitek:
Are you aware of https://github.com/zakodium-oss/react-science/issues/472?

Xstoudi:
> Are you aware of #472?

Yes, I just forgot to ref the issue.

Xstoudi:
Okay, keyboard navigation seems to be failing because `document.activeElement` stays the same even after changing focus.
[Select.tsx line 543](https://github.com/radix-ui/primitives/blob/eca6babd188df465f64f23f3584738b85dba610e/packages/react/select/src/Select.tsx#L543)

I'm not really sure what is causing this right now.

targos:
Probably related to the shadow dom. We often have issues with libraries that don't work well in a shadow dom.

Xstoudi:
I already had the problem with react libs to pinch/pan/zoom on elements. 

Radix seems to has some issues related to shadow dom:

https://github.com/radix-ui/primitives/issues/1980
https://github.com/radix-ui/primitives/issues/1772

Xstoudi:
Deploy successfull, I think you can review. If we merge like this I'll open an issue.

stropitek:
Behavior + API LGTM except for the issues you already mentioned.

Can you add an optional `style` prop (mainly to be flexible with the width of the select). I can merge after that.

I have a few comments on the visual look but will create separate issues for that.

targos:
There are linter issues btw.

Xstoudi:
I fixed linting errors and added the option to pass style props to the root element.

PR #497, opened 2023-06-07
fix: pass down style props on modals

Xstoudi:
The change allow user to decorate over already made components like `Modal.Body` or `Modal.Header`.

A workaround can be to re-encapsulate the `Modal.Header` in another `div` styled with what we want, but if we are working with heights it is not alway sufficient.


targos:
I'm a bit reluctant, because one of the goals of `react-science` is to help make coherent user interfaces. If you want to pass arbitrary root styles to components exported here, it seems to indicate that we have failed to make good enough components.

Issue #498, opened 2023-06-07
Make Select a controlled component


stropitek:
Did not notice this right away, but the component is not controlled. If `onSelect` is not passed to control some state outside of the component, it's not supposed have any effect when selecting an option.

`value` should be a mandatory prop, with the possibility to pass `undefined`

Xstoudi:
On Radix side, this is not how it's handled.

If value is undefined, then the component is considered as uncontrolled and manage it's own state with no regard of `onSelect` value.

I'll push more stories to allow you to be more precise on what behavior is not as intended.

stropitek:
Maybe I don't see the whole picture but IMO it can simplify the code in some cases but in general there are some downsides:
- It can lead to bugs where you think you have a controlled component but you forgot to pass the value and you end up with duplicated state.
- It becomes more code and work to make all components work in a coherent way (including components which don't use radix-ui)

I'm not completely opposed to it, but I haven't used uncontrolled components in a very long time and maybe I'm missing the upsides.



I would be in favour of fully controlled components just because it's an easier mental model if all of our components behave th

PR #501, opened 2023-06-07
chore: add more stories to demonstrate behavior

Xstoudi:
Refs: https://github.com/zakodium-oss/react-science/issues/498

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "@lukeed/uuid": "^2.0.1",
     "@popperjs/core": "^2.11.7",
     "@radix-ui/react-checkbox": "^1.0.4",
-    "@radix-ui/react-select": "^1.2.2",
+    "@radix-ui/react-select": "^2.0.0-rc.7",
     "@tanstack/react-query": "^4.29.5",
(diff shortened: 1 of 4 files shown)


stropitek:
Can you do an example of clearing the select value with a button both in the controlled vs uncontrolled cases? With the non-value being `undefined`?

package.json
```diff
@@ -67,7 +67,7 @@
     "@lukeed/uuid": "^2.0.1",
     "@popperjs/core": "^2.11.7",
     "@radix-ui/react-checkbox": "^1.0.4",
-    "@radix-ui/react-select": "^1.2.2",
+    "@radix-ui/react-select": "^2.0.0-rc.7",
```
targos, line 70: Are we ok with using a prerelease version? Why do you need it now?

stropitek: I asked @Xstoudi  to check if the behavior in v2 is still the same regarding a trick that was used to force radix-ui's select component to be controlled when the value is `undefined`.

But beyond this check we can go back to v1 to stay on the safe side.

stropitek (approved):

PR #507, opened 2023-06-19
fix: export MenuOption and MenuOptions types

Xstoudi:


stropitek:
Do you need it for the types or for the `MenuItems` component? `MenuItems` seems to be an internal component used by `DropdownMenu`, or am I wrong?

Xstoudi:
I want to type the variable I pass to `DropdownMenu`'s `options`.

I also can export `MenuOption` and `MenuOptions` without exporting whole `MenuItems`.

stropitek:
> I also can export `MenuOption` and `MenuOptions` without exporting whole `MenuItems`

Yes I think that would be better

src/components/dropdown-menu/index.ts
```diff
@@ -1 +1,2 @@
 export * from './DropdownMenu';
+export { MenuOptions, MenuOption, MenuDivider } from './MenuItems';
```
targos, line 2:
```suggestion
export type { MenuOptions, MenuOption, MenuDivider } from './MenuItems';
```

stropitek (approved):

PR #510, opened 2023-06-22
fix: modal generate a new portal context

Xstoudi:
Modal now create a root layout to ensure Select components opens in front of the modal instead of behind.


targos:
The menu opens correctly but on my browser (chrome based), it closes the modal immediately on click.

Probably need to port https://github.com/zakodium/components/pull/1123

targos:
You can start chromium with: `npx playwright cr`

Issue #471, opened 2023-04-05
Create textarea component


stropitek:
Should be designed very similarly to the input component in all its states (focused, hovered...)

Original source of inspiration for the inputs were from [ant.design](ant.design).

https://ant.design/components/input#components-input-demo-textarea

Stories to add
- Control story for the textarea component
- Story of a form with input and textarea components. Should look something like https://ant.design/components/form#components-form-demo-nest-messages. You can use / adapt the `Field` component for that.

PR #516, opened 2023-07-02
feat: implement textarea component

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/471

### DIFF
--- a/src/components/forms/TextArea.tsx
+++ b/src/components/forms/TextArea.tsx
@@ -1,5 +1,5 @@
 import styled from '@emotion/styled';
-import { CSSProperties, InputHTMLAttributes } from 'react';
+import { CSSProperties, TextareaHTMLAttributes } from 'react';
 
 import { FullSpinner } from '../index';
 
@@ -8,21 +8,23 @@ import {
   RenderAddon,
   RootInput,
   GroupStyled,
-  LeadingAddonStyled,
   LabelStyled,
+  LeadingAddonStyled,
   LeadingInlineAddonStyled,
   TrailingInlineAddonStyled,
   TrailingAddonStyled,
 } from './styles';
 
-const InputStyled = styled.input`
+const TextAreaStyled = styled.textarea`
   padding: 0;
   flex: 1 1 0%;
   border: none;
   position: relative;
   outline: none;
 `;
-export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
+
+export interface TextAreaProps
+  extends TextareaHTMLAttributes<HTMLTextAreaElement> {
   variant?: 'default' | 'small';
 
   leadingAddon?: RenderAddon;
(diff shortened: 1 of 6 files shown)


src/components/forms/TextArea.tsx
```diff
@@ -0,0 +1,126 @@
...
+  position: relative;
+  outline: none;
+`;
+
+export interface TextAreaProps
+  extends TextareaHTMLAttributes<HTMLTextAreaElement> {
+  variant?: 'default' | 'small';
+
+  leadingAddon?: RenderAddon;
+  trailingAddon?: RenderAddon;
```
stropitek, line 31: Those 2 props should not exist on the TextArea component. It doesn't really make sense to have them because the component expands vertically and it really doesn't look nice to have them / it wastes a lot of screen real estate.

stories/components/form.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+          onChange={(checked) =>
+            checked
+              ? setQuestion([...question, 3])
+              : setQuestion(question.filter((item) => item !== 3))
+          }
+        />
+      </Field>
+
+      <Field label="Introduction" name="introduction">
+        <TextArea maxLength={50} />
```
stropitek, line 81: I think we should avoid using `maxLength` in stories (in a textarea, it's confusing that typing in more than X characters stops working at some point).

stories/components/textarea.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  return (
+    <ExampleGroup style={{ width: '80px' }}>
+      <TextArea {...prop} value={state} onChange={onChange} />
+    </ExampleGroup>
+  );
+}
+Control.args = {
+  placeholder: 'Control example',
+  maxLength: 10,
```
stropitek, line 39: ditto

src/components/forms/styles.ts
```diff
@@ -0,0 +1,142 @@
...
+  padding-right: 0.5rem;
+`;
+
+export const TrailingInlineAddonStyled = styled.div`
+  display: flex;
+  align-items: center;
+  padding-left: 0.5rem;
+`;
+
+export const RootInput = styled.div`
```
stropitek, line 133: Can you rename to something else? `RootInput` suggests it's an input but it's not.

Maybe `InputWrapper` or `InputContainer`

src/components/forms/TextArea.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  help?: string;
+  error?: string;
+  valid?: true | string;
+
+  loading?: boolean;
```
stropitek, line 37: Loading can also be removed. Could be potentially useful but would need to be designed differently and we don't have a use case for now.

stropitek (commented):
CleanShot 2023-08-04 at 10 27 54

Help, error and valid text wraps too early

wadjih-bencheikh18:
@stropitek all the points you mentioned have been resolved!

stropitek:
- Resolve conflicts
- Remove loading from textarea controls
- Help / error / valid still has the same wrapping problem, I does not seem to be fixed

stropitek (approved):

PR #518, opened 2023-07-13
fix: allow to pass tag type for dropdown menu button

Xstoudi:
I'm not really sure if it's the best way to do. At some point it could be nice to switch from headless to radix as the component could benefit from the `asChild` prop of trigger's.


stropitek:
@Xstoudi I'm closing this since it's not relevant anymore as we are switching to blueprintjs library

If there is a use case this PR was fulfilling, feel free to open an issue.

PR #523, opened 2023-07-30
Implement Panelheader

### LINKED ISSUE #519: New component: PanelHeader
A PanelHeader could be used in the panel as a fixed header. It has 3 props:

- children
- count - if defined we put the number in square bracked on the right
- onClickSettings - if defined we add the gear on the right

It should replace in NMRium all the header of panels:

image

It should also be used in Pixelium

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/519


lpatiny:
Looks good to me. 2 remarks:

image

1. I think the icons should be closer to each other like in NMRium
2. Should the thin layer under the button not be included in this component ?

image

lpatiny:
I also found out that sometimes we have the current number and total number.

Please rename count to `total` and a a new property `current`.

If `current` and `total` are defined you could display 1 / 5 and keep the square brackets only when one of the 2 values are defined..

wadjih-bencheikh18:
Nmrium:
image

Our component:
image

lpatiny (approved):
Functionality looks ok to me

Issue #520, opened 2023-07-27
New component: PanelPreferencesToobar


lpatiny:
We need a component that would display preferences toolbar and would also have the validate / cancel icons.

2 props:

- onValidate
- onCancel

![image](https://github.com/zakodium-oss/react-science/assets/1484241/275e4b3f-a5bb-4a28-af14-8e7d552a5f1c)

PR #524, opened 2023-07-30
Implement PanelPreferencesToolbar component

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/520

### DIFF
--- a/src/components/toolbar/PanelPreferencesToolbar.tsx
+++ b/src/components/toolbar/PanelPreferencesToolbar.tsx
@@ -1,0 +1,49 @@
+/** @jsxImportSource @emotion/react */
+import { css } from '@emotion/react';
+import { ReactNode } from 'react';
+import { FaCheck, FaTimes } from 'react-icons/fa';
+
+import { Button } from '../button/Button';
+
+interface PanelPreferencesToolbarProps {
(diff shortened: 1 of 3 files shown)


lpatiny:
I think the buttons should be closer to each other like in NMRium

image

wadjih-bencheikh18:
Nmrium:
image

Our component:
image

The two are size `16px`

src/components/toolbar/PanelPreferencesToolbar.tsx
```diff
@@ -0,0 +1,55 @@
+/** @jsxImportSource @emotion/react */
+import { css } from '@emotion/react';
+import { ReactNode } from 'react';
+import { FaCheck, FaTimes } from 'react-icons/fa';
+
+import { Button } from '../button/Button';
+
+interface PanelPreferencesToolbarProps {
+  children?: ReactNode;
```
targos, line 9: The toolbar should not accept children, and not render the entire panel container. Just the toolbar, like in your other PR.

wadjih-bencheikh18: okey, fixed

Issue #533, opened 2023-08-09
View stories' source code


stropitek:
By installing the `@storybook/addon-storysource` plugin

PR #534, opened 2023-08-09
View stories' source code

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/533

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,6 +103,7 @@
     "@playwright/experimental-ct-react": "^1.36.2",
     "@playwright/test": "^1.36.2",
     "@storybook/addon-essentials": "7.2.1",
+    "@storybook/addon-storysource": "7.2.1",
     "@storybook/blocks": "7.2.1",
     "@storybook/react": "7.2.1",
     "@storybook/react-vite": "7.2.1",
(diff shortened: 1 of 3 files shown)


wadjih-bencheikh18:
the webkit test is not stable (It works in local after updating playwright)
should we add retries in playwright?

targos:
> the webkit test is not stable (It works in local after updating playwright) should we add retries in playwright?

It would be nice if we can understand and fix the root cause. As a last resort, we would add retries.

I think the first thing to do should be to upload playwright reports when tests fail, like we do already in cheminfo/nmrium

package.json
```diff
@@ -103,6 +103,7 @@
     "@playwright/experimental-ct-react": "^1.36.2",
     "@playwright/test": "^1.36.2",
     "@storybook/addon-essentials": "7.2.1",
+    "@storybook/addon-storysource": "^7.2.2",
```
targos, line 106:
```suggestion
    "@storybook/addon-storysource": "7.2.1",
```

Please revert the changes and redo with a fixed version (same for all storybook packages). Storybook doesn't like it when we mix up package versions.

targos (approved):

PR #535, opened 2023-08-10
Cleanup playwright test

wadjih-bencheikh18:
I found out that `dragTo` function is not stable in our case
```tsx
await component.dragTo(component, {
  sourcePosition: { x: 100, y: 100 },
  targetPosition: { x: 200, y: 200 },
}); 
```
So I changed it with manual dragging [playwright_documentation](https://playwright.dev/docs/input#dragging-manually)
```tsx
await component.hover({ position: { x: 100, y: 100 } });
await page.mouse.down();
await component.hover({ position: { x: 200, y: 200 } });
await page.mouse.up();
```

And I checked other tests and did a cleanup


stropitek:
Nice that those tests are finally passing!

stropitek:
Can't really review because the renaming of files makes it hard to see what changed, but I trust you on this one.

Issue #474, opened 2023-04-05
Create radio component


stropitek:
## Features
- Controlled component
- Classic radio or button styles
- Disabled

## API
```tsx
interface ValueLabel {
  value: string;
  label: ReactNode;
  disabled?: boolean;
}
const options: ValueLabel[] = [
  { label: 'Option 1', value: 'option1' },
  { label: 'Option 2', value: 'option2', disabled: true },
  { label: 'Option 3', value: 'option3' },
];

function Radios() {
  return (
    <>
      <Radio
        selected={options[0]}
        style="classic"
        options={options}
        onSelect={(option: ValueLabel) => {
          // set some state
        }}
      />
      <Radio style="button" options={options} />
    </>
  );
}

```

## Design

Something similar to the ant design: https://ant.design/components/radio

For the button style, use the outlined version (not the solid one).

![CleanShot 2023-04-05 at 11 09 54](https://user-images.githubusercontent.com/4118690/230037634-8ed5ffb5-ccd2-4d0f-aa9b-0bf44b303dfa.png)

## Requirements
Use radix-ui radio group component: https://www.radix-ui.com/docs/primitives/components/radio-group#radio-group

PR #536, opened 2023-08-11
Implement Radio Component

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/474


stropitek (commented):
Thank you, the implementation SGTM. I'd like to have a few design improvements and api additions before we merge this.

CleanShot 2023-08-11 at 10 05 53

Given that we are here implementing the "normal" variant of the component, can you make it look like the middle one? Can you also implement the `variant="small"` prop?

CleanShot 2023-08-11 at 10 11 23

Can you make it such that we don't see the grey outline when the radio is selected?


https://github.com/zakodium-oss/react-science/assets/4118690/8d725758-fa74-4e95-a064-b6ca4cda0cc5

Can you change the color of the outline when the mouse hovers the radio so that it matches the color of a selected radio?

CleanShot 2023-08-11 at 10 14 19

Can you avoid the double border on the radio groups.

Could you also show what happens when there is only one option in the select in the stories?

Can you add the `disabled` property to the `RadioGroup` so that it disables the entire component?

wadjih-bencheikh18:
> Can you avoid the double border on the radio groups.

I don't see this in my browser could you recheck again please

wadjih-bencheikh18:
@stropitek  I did all the improvement you asked thank you

src/components/forms/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { css } from '@emotion/react';
+import * as RadioGroupRadix from '@radix-ui/react-radio-group';
+import { ReactNode } from 'react';
+
+import { InputVariant } from '../styles';
+
+import { RadioOptionButton } from './RadioOptionButton';
+import { RadioOptionClassic } from './RadioOptionClassic';
+
+export interface ValueLabel {
```
stropitek, line 11: Please rename to `RadioOption`

src/components/forms/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,97 @@
...
+  label: ReactNode;
+  disabled?: boolean;
+}
+export interface RadioGroupProps {
+  selected?: ValueLabel;
+  type?: 'classic' | 'button';
+  options?: ValueLabel[];
+  onSelect?: (option: ValueLabel) => void;
+  name?: string;
+  disabled?: boolean;
```
stropitek, line 22: Can you add a story which shows how this prop works?

stropitek:
> I don't see this in my browser could you recheck again please

To see the problem, don't use the zoom of the browser, use the zoom of storybook.

If you look closely, you'll see the issue at 100% zoom as well.

wadjih-bencheikh18:
> If you look closely, you'll see the issue at 100% zoom as well.

I am not seeing the double border. Here's my result:
image

Could you please let me know which web browser you're using?

src/components/forms/RadioGroup/RadioGroup.tsx
wadjih-bencheikh18: Do you want me to rename the file? or the folder?

targos:
I see a double border with Chrome: 
CleanShot 2023-08-14 at 12 30 09

wadjih-bencheikh18:
> I see a double border with Chrome:

I checked safari version and yes it have double border
I fixed it now, Could you please check?

wadjih-bencheikh18:
> Did you also check on the deployed storybook (https://474-create-radio-component.react-science.pages.dev/) or only locally?

I checked the two

wadjih-bencheikh18:
> Safari is also not working for me. Did you also check on the deployed storybook (https://474-create-radio-component.react-science.pages.dev/) or only locally?

I found that navigators does not accept border less then 1px so I refactored the code to avoid it

stories/components/radio.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+        onSelect={setOption}
+        {...props}
+      />
+    </ExampleGroup>
+  );
+}
+Control.args = {
+  variant: 'default',
+  type: 'classic',
+  disabled: true,
```
stropitek, line 76:
```suggestion
  disabled: false,
```

stropitek:
The borders look good for me now.

> CleanShot 2023-08-11 at 10 05 53
> 
> Given that we are here implementing the "normal" variant of the component, can you make it look like the middle one? Can you also implement the `variant="small"` prop?

Can you make the button version more like the ones highlighted in the screenshot (less padding)

src/components/forms/RadioGroup/RadioGroup.tsx
stropitek: Juste the interface

wadjih-bencheikh18:
> Can you make the button version more like the ones highlighted in the screenshot (less padding)

Style updated.

stropitek:
I think font size should be 12px both on button and classic versions of the small variant

Currently classic is 12px but button is 14px

wadjih-bencheikh18:
> Currently classic is 12px but button is 14px

It's fixed now

I used 14 in button style in two variant because it's the case in the ant design

stropitek (approved):

Issue #547, opened 2023-08-24
Extend the input component to have a 'search' component


lpatiny:
When there is a value the 'x' allows to clear the value.

There is a placeholder
There is a defualt 'magnifier' in front.

![2023-08-24 12 21 25](https://github.com/zakodium-oss/react-science/assets/1484241/d0bc01ab-b4ee-4a77-aa07-6fca64f2972d)

Once done this should be:
- implemented in the InfoPanel
<img width="531" alt="image" src="https://github.com/zakodium-oss/react-science/assets/1484241/b00fffc2-f830-414d-b8e6-c5a9fabf35b0">
- In NMRium database search
<img width="1058" alt="image" src="https://github.com/zakodium-oss/react-science/assets/1484241/f0a38491-0f28-4098-9b1a-7013f42b5355">

stropitek:
Placeholder and "inline leading addon" (for the search icon) are already implemented.

To implement: add `onClear` prop which takes a callback and adds a clear button which calls the callback if the prop is defined.

IMO we can make the clear button more subtle (no background color) than what we have in nmrium.

stropitek:
Passing `type="search"` when used in blueprintjs's `InputGroup` will format the input slightly differently and use the native "clear" feature. It's not supported in firefox (currently behind a flag with [an issue open](https://bugzilla.mozilla.org/show_bug.cgi?id=1654288) to have it the default).

https://blueprintjs.com/docs/#core/components/input-group.search-input

I think we can work with that.

Would be good though to have a story which shows the component in action, and shows what happens with `onValueChange` / `onChange` when the input is cleared.

Issue #521, opened 2023-07-27
`ValueRenderers.Number` to render nothing if passed `value` is falsy.


Xstoudi:
`0` being falsy, it leads `ValueRenderers.Number` to render nothing if `value` is zero.
https://github.com/zakodium-oss/react-science/blob/main/src/components/value-renderers/Number.tsx#L10

PR #548, opened 2023-08-27
547 extend the input component to have a search component

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/547

### DIFF
--- a/src/components/info-panel/InfoPanel.tsx
+++ b/src/components/info-panel/InfoPanel.tsx
@@ -135,6 +135,20 @@ export function InfoPanel(props: InfoPanelProps) {
           if (target.value !== undefined) setSearch(target.value);
         }}
         style={inputStyle}
+        leadingAddon={{ addon: <FaSearch />, inline: true }}
+        trailingAddon={
+          search
+            ? {
+                addon: (
+                  <FaTimesCircle
+                    onClick={() => setSearch('')}
+                    style={{ cursor: 'pointer' }}
+                  />
+                ),
+                inline: true,
+              }
+            : undefined
+        }
       />
 
       {isArray(dataArray)
--- a/src/components/info-panel/InfoPanel.tsx
+++ b/src/components/info-panel/InfoPanel.tsx
@@ -1,10 +1,9 @@
 /** @jsxImportSource @emotion/react */
 import { css } from '@emotion/react';
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment, @typescript-eslint/prefer-ts-expect-error
-// @ts-ignore
 import { Disclosure } from '@headlessui/react';
-import { CSSProperties, useState } from 'react';
-import { FaChevronRight } from 'react-icons/fa';
+import { isArray } from 'lodash';
+import { CSSProperties, useCallback, useState } from 'react';
+import { FaChevronRight, FaSearch, FaTimesCircle } from 'react-icons/fa';
 
 import { Input, ValueRenderers } from '../index';
 import { Table } from '../table/Table';
@@ -16,7 +15,7 @@ export interface InfoPanelData {
 
 interface InfoPanelProps {
   title?: string;
-  data?: InfoPanelData[];
+  data?: InfoPanelData[] | InfoPanelData;
   titleStyle?: CSSProperties;
   inputStyle?: CSSProperties;
   tableStyle?: CSSProperties;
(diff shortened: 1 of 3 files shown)


stropitek:
@wadjih-bencheikh18 you haven't implemented the "clearable input" feature as described in  https://github.com/zakodium-oss/react-science/issues/547

The idea was to add a `clearable: boolean` prop to the `Input` component.

Please make a story and handle the edge case "the input is clearable and has a trailing addon"

src/components/forms/Input.tsx
```diff
@@ -88,14 +97,31 @@ export function Input(props: InputProps) {
...
+      const event = {
+        target: ref.current,
+      } as ChangeEvent<HTMLInputElement>;
+      onChange?.(event);
+    }
+  };
+  const [value, setValue] = useState(otherProps.value);
+  useEffect(() => {
+    setValue(otherProps.value);
+  }, [otherProps.value]);
```
stropitek, line 124: The component is controlled and should not have this kind of internal state or a `useEffect`

I think the problem your are running into is the `event` param of `onChange`, which does not play nice with this new `clearable` option.

I've done a little bit of research, and ant has a similar api, and they use a library which solves it like this: https://github.com/react-component/input/blob/aa08e91ce278103f3d0b5d1642ee541fe87e7ad2/src/utils/commonUtils.ts#L12

I don't like this solution. It pretends that the event originates from the input when it actually comes from the clear button.

My suggestion is to make a breaking change:
`onChange` -> `onValueChange`
`onChange` cannot be passed down from the `Input` to the `input`. It is created internally and forwards events to `onValueChange`.

```ts
interface InputProps {
  // The second argument is passed only when the change in value
  // comes from the input's onChange event
  onValueChange: (value: string, event?: React.ChangeEvent<HTMLInputElement>);
}

@targos what do you think?

stropitek:
You you create a controlled story for `Input`?

src/components/forms/Input.tsx
targos: Shouldn't we just change the value and use `dispatchEvent` to dispatch the change event?

targos: I also read in many places that we should dispatch an `input` event: https://stackoverflow.com/questions/23892547/what-is-the-best-way-to-trigger-change-or-input-event-in-react-js

stropitek: > Shouldn't we just change the value and use `dispatchEvent` to dispatch the change event?

IDK, really not sure about what to do with this one. By doing this you can end up in a situation where the `value` prop of the `Input` and the actual value of the `input` are not synced. Is it acceptable?

targos: Isn't it what happens already with React when a component is controlled? When you listen to the change event, `event.target.value` is already changed.

stropitek: Yes it's true, haven't thought about it that way.

@wadjih-bencheikh18 can you give the `dispatchEvent` solution a shot? Normally in the end you should be able to get rid of the `useState` and `useEffect` in your implementation.

wadjih-bencheikh18: > @wadjih-bencheikh18 can you give the `dispatchEvent` solution a shot?

Okay, I will try it

The reason why I used the `useEffect` and `useState` was the case where the `value` and `onChange` are not defined which I don't think `dispatchEvent` will solve it

Here's an example
https://github.com/zakodium-oss/react-science/blob/b66969a61a2ad6e946e6fbe264aeb2ed4907dcee/stories/components/input.stories.tsx#L85-L88

stropitek: ok i see

In that case I think we can have 2 controlled stories. One for controlled input and one for the uncontrolled input.

stropitek: I'm not sure if having an uncontrolled version of the input which re-renders on every value change because it has internal state tied to it really makes sense.

Maybe the `clearable` prop should only work for the controlled input and we log a warning when it's used inappropriately.

targos: But I'm also expecting it to work when it's uncontrolled. There are two steps in the solution I proposed:
- change the value programmatically
- create and send the event

wadjih-bencheikh18: Using `dispatchEvent` did not update the state in case of controlled input
likely due to the issue discussed here: https://stackoverflow.com/a/39066491

wadjih-bencheikh18:
> @wadjih-bencheikh18 can you create another PR which only fixes #521 ?
> It's already fixed by https://github.com/zakodium-oss/react-science/pull/566 we can close it

PR #555, opened 2023-09-26
feat: remove shadow dom and global styles

stropitek:
instruct user how to add the proper global styles


stropitek:
For the commit message when merging:

BREAKING CHANGE: the shadow dom has been removed from the root layout and global styles no longer apply to it. Global styles are published by the library so that users can use them in their applications.

targos:
Changes LGTM, but let me try with a prerelease on nmrium.

PR #557, opened 2023-09-27
feat: improve splitter component to work on touch devices 

hamed-musallam:
- Adjusting the size of the splitter by touch and move. 
- Close the splitter by touching it with two fingers.


targos:
I think the first commit is non-controversial. Can you remove the other one so we can merge?

hamed-musallam:
done, can you merge it?

PR #559, opened 2023-10-03
refactor!: migrate Toolbar with blueprintjs components

stropitek:
- chore: apply blueprintjs css global stylesheets
- refactor: use blueprintjs components in Toolbar component


stropitek:
There is a story to see the toolbar with controls:
https://blueprintjs.react-science.pages.dev/stories/?path=/story/components-toolbar--control

PR #568, opened 2023-10-17
chore: remove commonjs build and fix esm build

stropitek:


stropitek:
Just read this [here](https://www.totaltypescript.com/concepts/option-module-must-be-set-to-nodenext-when-option-moduleresolution-is-set-to-nodenext#youre-using-the-wrong-moduleresolution)

CleanShot 2023-10-17 at 14 52 12

targos (approved):

Issue #571, opened 2023-10-18
Rewrite buttons to be wrappers around blueprintjs's buttons


stropitek:
https://blueprintjs.com/docs/#core/components/buttons

Here I think we can add a bit of sugar on top of the blueprint API.

For the `Button` component:

We can make it easier to add a tooltip to a Button than would be for any other component, because it's a frequent use case.

But mostly the component should just forward props to the blueprintjs component.

## API suggestion

```ts
type ButtonProps = BlueprintButtonProps & {
  tooltipProps: Omit<BlueprintTooltipProps, 'children'> 
}
/>
```

There is a known limitation when using a tooltip on a `Button`, it will not work if the button is disabled. Therefore I suggest that we render a blueprintjs `AnchorButton` when disabled, so that the tooltip still shows up.

wadjih-bencheikh18:
I don't like that in `BlueprintButtonProps` there is two props `large` and `small` in order to change button size. And also It's unstable if we toggle them randomly.
I suggest to use one prop `size: 'regular' | 'small' | 'large'`

stropitek:
> I suggest to use one prop `size: 'regular' | 'small' | 'large'`

I agree in general, but I think since the goal of this component is to add something on top of the base bluprintjs button, not change / improve its api.

When building applications with react-science, we will import re-usable components with which are either:
- directly imported from `blueprintjs`. For example `ContextMenu` which also has a `larger` prop.
- imported from `react-science`
  - Component built on top of blueprintjs like `Button`, should share api as much as possible
  - Standalone component like `SplitPane`


> And also It's unstable if we toggle them randomly

What do you mean? It produces strange results when you activate both props together?

wadjih-bencheikh18:
> It produces strange results when you activate both props together?

Yes

wadjih-bencheikh18:
> Therefore I suggest that we render a blueprintjs `AnchorButton` when disabled, so that the tooltip still shows up.

There's a diffrance between `BlueprintButtonProps` and `BlueprintAnchorButtonProps` types. Do you suggest a solution in order to keep possible props to our component.

stropitek:
AFAIK the difference in the props only come down to the differences between the attributes a `<button>` and an `<a>` element can take. Maybe you can use the intersection between `button` and `a` attributes and only allow those?

If that does not work you can only spread props when the button is rendered but for the anchor pass hand-picked props?

PR #576, opened 2023-11-01
571 rewrite buttons to be wrappers around blueprintjss buttons

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/571


src/components/button/Button.tsx
```diff
@@ -1,66 +1,31 @@
...
-        },
-      })}
-      type="button"
-      disabled={disabled}
-    >
-      {children}
-    </button>
+    <InnerButton {...buttonProps}>
+      <Tooltip {...tooltipProps}>{children}</Tooltip>
+    </InnerButton>
```
stropitek, line 29: Check that the tooltip is really activated on the whole button, also when bringing the cursor close to the edges.

stropitek (changes requested):
CleanShot 2023-11-02 at 11 30 17

Can you add some padding so that we can see the border of the button better in the button stories?

---

Can you add a story which is button with a tooltip in which the tooltip props are controllable?

---

I think the disabled story can be removed since that can be controlled in the main controllable story.

---

There is a regression in the toolbar buttons: https://react-science.pages.dev/stories/?path=/story/components-aboutdialogtoolbarbutton--display-about-dialog

On main: 
CleanShot 2023-11-02 at 11 32 37

On this PR: 
CleanShot 2023-11-02 at 11 32 47

---

There is also a regression here (PanelPreferencesToolbar story). I think the minimal version of the button with the icon prop (use blueprint icons) should work no?

Other regression: PanelHeader

---

Can the close button of the modal also be written with the button component? Again always try to use icons from the bluprintjs library first.

src/components/button/Button.tsx
wadjih-bencheikh18: resolved

wadjih-bencheikh18:
> use blueprint icons

I used blueprint icons in all cases, but we can see that these icons are grey
Do I have to change the color or keep it like that (I prefer the actual colors)

src/components/button/Button.tsx
```diff
@@ -1,66 +1,31 @@
 /** @jsxImportSource @emotion/react */
```
stropitek, line 1: This line can be removed

stropitek (approved):
Can you check why the `fill` prop has no effect in the Button story?

It's supposed to be working like here: https://blueprintjs.com/docs/#core/components/buttons

Otherwise LGTM

wadjih-bencheikh18:
> Can you check why the `fill` prop has no effect in the Button story?

The problem is happening because `button` is wrapped inside the `tooltip` so he's filling the tooltip.
But I fixed it.

PR #579, opened 2023-11-10
Use blueprintjs InputGroup and TextArea

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/547

PR #584, opened 2023-11-21
fix TS esm build and storybook build

stropitek:
- chore: fix ts build error
- chore: fix webpack issues


stropitek:
@targos this is to fix the regression introduced by https://github.com/zakodium-oss/react-science/commit/a85d339d06aae45371349ec7ecb2607df25ee61a

@wadjih-bencheikh18 the problems you have with the build in https://github.com/zakodium-oss/react-science/pull/579 should be fixed by this.

targos (approved):

PR #586, opened 2023-12-03
Use NonIdealState for empty DropZone

### LINKED ISSUE #581: Reword design of dropzone component
The ["non-ideal state" component](https://blueprintjs.com/docs/#core/components/non-ideal-state) matches what we want in a dropzone component when nothing is loaded.

We can update the DropZone component and [this story](https://react-science.pages.dev/stories/?path=/story/components-dropzone--control) to use the nicer design.

Suggested API:

```jsx
<DropZone  
  emptyIcon="import"
  emptyTitle="No data loaded"
  emptyDescription="You can load data by drag-and-dropping files here"
  emptyButtonText="Select files"
  emptyButtonIcon="plus"
/>
```

Some properties are optional as in the `NonIdealState` component.

The `color` prop can be removed.

moonayyur:
closes #581

### DIFF
--- a/src/components/drop-zone/DropZone.tsx
+++ b/src/components/drop-zone/DropZone.tsx
@@ -35,7 +35,7 @@ const DropzoneDragActive = styled.div<DropzoneColorProps>`
   bottom: 0;
   left: 0;
   right: 0;
-  background-color: white;
+  background-color: rgb(255, 255, 255, 0.7);
   border: 5px dashed;
   display: flex;
   flex-direction: column;


stropitek:
Also the style (font size, weight etc.) should be more consistent between when there is a drop target and when there is none:


https://github.com/zakodium-oss/react-science/assets/4118690/4dce2015-c859-41fc-9c83-79dd43d2a604

stropitek:
What happens when the dropzone has children has slightly changed. Before the children were still visible below an overlay.

Is it on purpose?

Before:

https://github.com/zakodium-oss/react-science/assets/4118690/d494c51a-4b14-40c8-b496-14856dfd62d2

After:

https://github.com/zakodium-oss/react-science/assets/4118690/39f1113f-f4a7-47f7-ac87-c4f18d482c85

PR #588, opened 2023-12-03
Refactor select using blueprintjs

### LINKED ISSUE #578: Replace Select component with blueprintjs component
@wadjih-bencheikh18 did you replace this component in NMRium?

We won't be able to match the features of the current component with the features of the Blueprintjs component.

You can remove the stories which can't be reproduced with the new `Select` component from blueprintjs.

Please add a story with a native `<select>`, I'd like to see if blueprint adds some styling to them and what it looks like.

wadjih-bencheikh18:
closes #578


stropitek (commented):
@wadjih-bencheikh18 

I made a few changes:
- Use of decorator to avoid repeating the same "shell" around each select story.
- Since the arrow was not placed correctly, I used the `minimal` popover prop. I gave it a quick try but I wasn't able to place the arrow properly. I don't know why but it behaves differently in the stories than in the blueprintjs documentation, where the arrow is centered with respect to the select button.

CleanShot 2023-12-04 at 11 28 42

Can you maybe have another look and try to find out why that is the case?


---

CleanShot 2023-12-04 at 11 30 44

The styles for the disabled select don't seem to be applied, it look exactly like the non-disabled one.

---

The font size in the list seems to be different when we are in a modal vs otherwise.

wadjih-bencheikh18:
> The font size in the list seems to be different when we are in a modal vs otherwise.

The storybook div have style `font-size:12px` and in the default case the `Select` popup use portal out of the storybook div 
image
So the font is 14px

In the `InModal` Story I didn't use the default Portal because the popup must be in the `dialog` otherwise it's blured, so I used the modal portal from the `rootLayoutContext` which make it inside the storybook div

stropitek (approved):

PR #589, opened 2023-12-04
fix: render explicit string literals for representing boolean values

### LINKED ISSUE #2784: spectra information panel not rendering the boolean object correctly in the table
Screenshot 2023-12-03 at 19 40 37

hamed-musallam:


stropitek (approved):

src/components/info-panel/InfoPanel.tsx
```diff
@@ -133,7 +133,7 @@ export function InfoPanel(props: InfoPanelProps) {
 function valueCell(value: number | string | object | boolean) {
   switch (typeof value) {
     case 'boolean':
-      return <ValueRenderers.Boolean value={value} />;
+      return <ValueRenderers.Boolean renderValue={String} value={value} />;
```
lpatiny, line 136: Why not just calling a Text renderer with String(value) ? I don't see why we should change the Boolean renderer.

hamed-musallam: we did this to render the value as a string "true" or "false" instead of a symbol

targos: @lpatiny's suggestion would give the same result.

hamed-musallam: The purpose is to provide users of the ValueRenderers.Boolean component with the flexibility to customize the rendering of strings, allowing them to choose representations that best convey true or false values according to their preferences

stropitek:
We discussed this and agreed that showing `Yes` / `No` as values would be a better default than `true` / `false`

PR #599, opened 2023-12-12
fix: toolbar item width

hamed-musallam:
we try to fix the issue of the toolbar item width 

Screenshot 2023-12-12 at 15 04 45
Screenshot 2023-12-12 at 14 57 37

### DIFF
--- a/src/components/toolbar/Toolbar.tsx
+++ b/src/components/toolbar/Toolbar.tsx
@@ -115,7 +115,7 @@ Toolbar.Item = function ToolbarItem(props: ToolbarItemProps) {
         }
       `}
       intent={intent}
-      style={{ position: 'relative', fontSize: '1.25em', width: 'auto' }}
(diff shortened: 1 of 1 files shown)


stropitek (approved):
What browser is affected by this?

It seemed to be working fine with in Chrome here: https://fix-wrap-button-with-span.react-science.pages.dev/stories/?path=/story/components-toolbar--vertical-toolbar

hamed-musallam:
the problem exists across all browsers. In the first image from NMRium the width toolbar items are set to 'auto,' and the container is a div styling with a flexbox


image

The second issue arises when the default width for the button is maintained as 'width: 100%', and the item is wrapped

image

hamed-musallam:
Now, when we apply 'width: fit-content' to the toolbar item, we achieve the desired style

https://update-react-science.nmrium.pages.dev/#/SamplesDashboard/7kn5wec7isa/Viewdatacytisinedatacytisine1Hjson

PR #615, opened 2023-12-18
feat: add a 'disabled' option to Toolbar.Item

hamed-musallam:
refactor: `intent` and `disabled` props at toolbar level will be overridden with the item props once it is defined
 at item level

### DIFF
--- a/src/components/toolbar/Toolbar.tsx
+++ b/src/components/toolbar/Toolbar.tsx
@@ -17,11 +17,13 @@ import {
   useToolbarContext,
 } from './toolbarContext';
 
-export interface ToolbarProps {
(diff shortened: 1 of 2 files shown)


stropitek (commented):
Can you edit the story which overrides the intent and do the same with disabled?

https://react-science.pages.dev/stories/?path=/story/components-toolbar--horizontal-toolbar

hamed-musallam:
> Can you edit the story which overrides the intent and do the same with disabled?
> 
> https://react-science.pages.dev/stories/?path=/story/components-toolbar--horizontal-toolbar

it is done

stropitek (approved):

PR #619, opened 2023-12-22
Only keep controls that affect tooltip design

### LINKED ISSUE #613: Tooltip button controls do not all work
https://react-science.pages.dev/stories/?path=/story/components-button--tooltip&args=intent:primary

Only keep the controls that are supposed to affect the tooltip visually, like intent. The other controls can be tested in the main Button story.

moonayyur:
closes #613

### DIFF
--- a/stories/components/button.stories.tsx
+++ b/stories/components/button.stories.tsx
@@ -64,6 +64,18 @@ Tooltip.argTypes = {
     options: ['none', 'primary', 'success', 'warning', 'danger'],
   },
 };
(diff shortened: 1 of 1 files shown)


stropitek:
Can you add a similar story on a disabled button?

It's a bit of an edge case we should be able to check because disabled buttons render a different component than regular buttons (see Button implementation)

stropitek:
ping @moonayyur

stropitek (approved):

PR #624, opened 2024-01-03
Remove Tabs component 

### LINKED ISSUE #573: Replace Tabs component
Replace usage with blueprintjs
https://blueprintjs.com/docs/#core/components/tabs

I think in this case we can remove the stories.

CleanShot 2023-10-18 at 11 55 04

to become

CleanShot 2023-10-18 at 12 01 08

Since there is built-in support for "tags", we can use that to show the number of spectra loaded for each type.

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/573


stropitek (commented):
CleanShot 2024-01-04 at 10 40 33
Also what about using a Badge to show the number of spectra of each type as suggested in the issue?

You can remove the "Controlled" story in tabs stories.

stropitek (approved):

PR #625, opened 2024-01-03
chore: upload playwright report when playwright tests fail

stropitek:


stropitek:
It seems like the tests are passing now that the trace option has been activated... 🤷🏻‍♂️

.github/workflows/playwright.yml
```diff
@@ -25,3 +26,10 @@ jobs:
         run: npx playwright install --with-deps
       - name: Run ${{ matrix.test-type }} tests on ${{ matrix.project }}
         run: npm run playwright-${{ matrix.test-type }} -- --project ${{ matrix.project }}
+      - name: Upload playwright report
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: playwright-report-${{ matrix.project }}-${{ matrix.test-type }}
+          path: playwright-report/
```
targos, line 34: This seems wrong, given the new report configs.

playwright-ct.config.mts
```diff
@@ -3,5 +3,7 @@ import { defineConfig } from '@playwright/experimental-ct-react';
 import config from './playwright-base.config.mjs';
 
 config.testDir = './tests/components/';
-
+config.reporter = process.env.CI
+  ? 'html'
+  : [['html', { outputFolder: 'playwright-report-ct' }]];
```
stropitek, line 8: Dev: report to a specific directory so that `e2e` and `ct` don't overwrite each other (important for `npm run playwright`).
CI: each type of test runs independently so we output to the default folder.

.github/workflows/playwright.yml
stropitek: It works, see https://github.com/zakodium-oss/react-science/actions/runs/7399720269

playwright-ct.config.mts
targos: Can you still add the github reporter in CI?

targos (approved):

Issue #550, opened 2023-09-15
Rewrite Modal components using blueprintjs's Dialog


lpatiny:
![image](https://github.com/zakodium-oss/react-science/assets/1484241/a9faa71d-c697-45ae-8625-a06076bca522)

There are multiple issues with the ConfirmModal, including how colors are defined and the component's API.

We have to work on buttons before handling this.

We would like this before merging: https://github.com/cheminfo/nmrium/pull/2618

stropitek:
- Rewrite Modals which are not ConfirmModal with blueprintjs's Dialog
- Rename `ConfirmModal` to `ConfirmDialog`. It should use blueprintjs's Dialog. I think all of `Dialog`'s props can be forwarded by this component.

@wadjih-bencheikh18 If you notice any feature which were possible before but are lost when switching to blueprintjs, please write it down in the PR and we'll see case by case what we should do.

PR #626, opened 2024-01-04
Rewrite modal components using blueprintjss dialog

wadjih-bencheikh18:
Closes : https://github.com/zakodium-oss/react-science/issues/550


stropitek:
Can you resolve the conflicts?

stories/components/modal.stories.tsx
stropitek, line 1: Rename this to `dialog.stories.tsx`

Also change the titles of all the stories accordingly

stories/components/modal.stories.tsx
```diff
@@ -255,12 +238,21 @@ export function DynamicallySizedChildren() {
...
         Open editor modal
       </Button>
-      <Modal width={700} height={500} isOpen={isOpen} onRequestClose={close}>
-        <Modal.Header>Test OCL editor in modal</Modal.Header>
-        <Modal.Body>
+      <Dialog
+        title="Test OCL editor in modal"
+        isOpen={isOpen}
+        onClose={close}
+        style={{ width: 700, maxHeight: 500 }}
```
stropitek, line 245: is the `maxHeight` property really needed here?

stories/components/panel-header.stories.tsx
```diff
@@ -39,29 +35,25 @@ export function WithModal() {
...
       </PanelHeader>
-      <Modal isOpen={isOpen} onRequestClose={close} width={400}>
-        <Modal.Header>Settings</Modal.Header>
-        <Modal.Body>
+      <Dialog
+        title="Settings"
+        icon="cog"
+        isOpen={isOpen}
+        onClose={close}
+        style={{ width: 400 }}
```
stropitek, line 43: What is the width when you don't specify it?

stropitek (commented):
Before:
CleanShot 2024-01-04 at 17 28 05

Now:
CleanShot 2024-01-04 at 17 27 17

The layout is looking better before. If you do not use the `DialogBody` / `DialogFooter` components, you can probably have something like we had before. You can still probably use the background color the `DialogBody` uses by importing it from blueprintjs. The library exports the hexadecimal values of its color palette.

CleanShot 2024-01-04 at 17 45 07

In the complex content story, please try to make the content of the Dialog content overflow.
When the text overflows, only the body part should scroll and there is a prop `useOverflowScrollContainer ` on `DialogBody` that should help with this.

stropitek:
> In the complex content story, please try to make the content of the Dialog content overflow.
> When the text overflows, only the body part should scroll and there is a prop `useOverflowScrollContainer ` on `DialogBody` that should help with this.

Actually this example is a bit too specific since the part that should scroll is not the body of the modal, it's a subpart of it. But I did not understand why `overflow: visible` is set on the modal.

wadjih-bencheikh18:
I think it's better to use 
https://blueprintjs.com/docs/#core/components/dialog.multistep-dialog for complexe dialog stories

stories/components/modal.stories.tsx
wadjih-bencheikh18: No, I just used the same old stories values.

stories/components/panel-header.stories.tsx
wadjih-bencheikh18: it will take content width, so no need to add `style={{ width: 400 }}`

stropitek:
> I think it's better to use
> https://blueprintjs.com/docs/#core/components/dialog.multistep-dialog for complexe dialog stories

This is only for dialogs that have **sequential** steps, which I don't think we have anywhere and which is not the case for the general configuration dialog that NMRium has.

I think we have to try to use the generic modal to do the configuration dialog.

wadjih-bencheikh18:
> I think we have to try to use the generic modal to do the configuration dialog.

I tried to fix the layout, could you please take a look

stropitek:
https://github.com/zakodium-oss/react-science/assets/4118690/12fd3eb6-52a3-4265-9bd2-f1381d563554

The vertical menu on the left should not scroll with the main content. Only the text should scroll.

stropitek (approved):
LGTM after the 2 last comments are resolved

wadjih-bencheikh18:
@stropitek fixed

stropitek:
I still have to scrollbar (edge / chrome)

CleanShot 2024-01-09 at 16 58 04

Can you rename the remaining stories that use the "Modal" term?

CleanShot 2024-01-09 at 16 59 09

wadjih-bencheikh18:
@stropitek fixed, Thank you

Issue #627, opened 2024-01-04
Toolbar in panel header story is broken


stropitek:
https://react-science.pages.dev/stories/?path=/story/components-panelheader--basic

![CleanShot 2024-01-04 at 14 09 39](https://github.com/zakodium-oss/react-science/assets/4118690/ed50d689-a3a7-4e11-a3ff-29f75bc069c6)

PR #629, opened 2024-01-04
Toolbar in panel header story is broken

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/627

### DIFF
--- a/src/components/toolbar/Toolbar.tsx
+++ b/src/components/toolbar/Toolbar.tsx
@@ -145,13 +145,14 @@ Toolbar.Item = function ToolbarItem(props: ToolbarItemProps) {
         onClick?.(props);
       }}
       tooltipProps={
-        (!noTooltip && {
-          content: title,
-          placement: vertical ? 'right' : 'bottom',
-          intent,
-          compact: !large,
-        }) ||
-        undefined
+        noTooltip
+          ? undefined
+          : {
+              content: title,
+              placement: vertical ? 'right' : 'bottom',
+              intent,
+              compact: !large,
+            }
       }
       {...other}
     />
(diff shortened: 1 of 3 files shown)


stropitek (approved):

src/components/toolbar/Toolbar.tsx
```diff
@@ -39,6 +39,7 @@ export interface ToolbarItemProps extends ToolbarBaseProps {
   active?: boolean;
   onClick?: (item: ToolbarItemProps) => void;
   className?: string;
+  tooltip?: boolean;
```
stropitek, line 42: I approved the PR a bit too quickly...

Can you make this a `noTooltip`. We always try to name props so that the default is `false`
- It's more intuitive to use since not specifying it is the same as setting it to `false`
- It's nicer syntaxically to do `<Toolbar.Item noTooltip />` `<Toolbar.Item tooltip={false} />`

wadjih-bencheikh18: fixed, thank you for the explication.

src/components/toolbar/Toolbar.tsx
```diff
@@ -145,14 +145,13 @@ Toolbar.Item = function ToolbarItem(props: ToolbarItemProps) {
...
-              compact: !large,
-            }
-          : undefined
+        (!noTooltip && {
+          content: title,
+          placement: vertical ? 'right' : 'bottom',
+          intent,
+          compact: !large,
+        }) ||
+        undefined
```
stropitek, line 154: The ternary was easier to read

wadjih-bencheikh18: fixed

stropitek (approved):

PR #633, opened 2024-01-19
Story with accordion + panel toolbar + stacked panel

### LINKED ISSUE #616: Story with accordion + panel toolbar + stacked panel
Do another story similar to this one:
https://583-use-blueprintjs-htmltabl.react-science.pages.dev/stories/?path=/story/components-panelheader--with-accordion

Except when you click any button in the toolbar, it opens a panel stack:
https://blueprintjs.com/docs/#core/components/panel-stack2

- The panel stack should open on top of the toolbar and take the entire space available in the accordion item (red rectangle). CleanShot 2023-12-18 at 10 31 53
- There should be confirm / close buttons which close the panel in a header. The panel can look similar to this: https://583-use-blueprintjs-htmltabl.react-science.pages.dev/stories/?path=/story/components-panelpreferencestoolbar--basic

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/616


wadjih-bencheikh18:
I couldn't understand why there's this type error

```
interface PanelInfo {
  content: ReactNode;
}
```

The error : PanelInfo is not constraint of object

wadjih-bencheikh18:
> The error : PanelInfo is not constraint of object

Fixed, @stropitek

stropitek:
The result looks ok, but have you tried to enable the animation?

wadjih-bencheikh18:
> The result looks ok, but have you tried to enable the animation?

In the previous implentation the headers wasn't included in the panels so there was no animation because I was just opening one stuckPanel and close it so the stack has only one element max.

But now the initial panelheader is also a stackpanel and closing header is a part of the other panels, so now we have the animation because we toggle between stackpanels

stropitek:
The animation is strange and looks different than what's in the blueprint documentation site. The second panel is supposed to overlay over the first one, but without the first panel moving.

Also there are scroll bars appearing.

wadjih-bencheikh18:
@stropitek Fixed could you please check?

stropitek:
Can you put some kind of title here?

CleanShot 2024-01-30 at 13 45 26

The first animation is coming from the right, the the subsequent ones from the left. Can you make it such that it always comes from the right?


https://github.com/zakodium-oss/react-science/assets/4118690/82633c6a-42d6-4c9b-9e49-941fcdf4a582

src/components/toolbar/PanelPreferencesToolbar.tsx
```diff
@@ -4,16 +4,28 @@ import { css } from '@emotion/react';
...
+  title?: string;
   onClose?: () => void;
   onSave?: () => void;
 }
 
 const styles = {
+  container: css({
+    display: 'flex',
+    justifyContent: 'space-between',
+    borderBottom: '0.55px solid rgb(240, 240, 240)',
```
stropitek, line 16: `0.55px` seems like a magical value.

It doesn't work well with `1px`?

wadjih-bencheikh18: I need to try it. It was always like that. when I implemented [PanelPreferencesToolbar.tsx](https://github.com/zakodium-oss/react-science/pull/633/files/480380937f3f382b250ebe0f99f6d98bded13063..e5481059da6338c15928fee55d36e018300311dc#diff-0117685563b69abe2dd16abb4d63a1a9994eb1bd279242649fa7001668243a83) I took the design directly from `nmrium`

wadjih-bencheikh18:
> The first animation is coming from the right, the the subsequent ones from the left. Can you make it such that it always comes from the right?

For the animation, I tried to copy the exact example in blueprint js https://blueprintjs.com/docs/#core/components/panel-stack2 and I got the same problem. It may be because the storybook container or a bug. So I will try it first outside storybook

wadjih-bencheikh18:
i confirmed that the storybook container caused that, I tried it in the app. and works perfectly

stropitek:
I will merge this after you fixed the conflicts.

I already prepared https://github.com/zakodium-oss/react-science/issues/643 which is a continuation of this.

wadjih-bencheikh18:
@stropitek I used the uncontrolled PanelStack2

stropitek (approved):

Issue #638, opened 2024-01-29
Clicking outside to close the Color Picker is closing the modal 


hamed-musallam:
https://react-science.pages.dev/stories/?path=/story/components-color-pickers--color-picker-dropdown-story

https://github.com/zakodium-oss/react-science/assets/35760236/444a83c8-5491-4e92-8920-0c56896ffebf

hamed-musallam:
After migrating to react-science version 0.33 and replacing the old Modal component with the Dialog component from Blueprint.js, the color picker is now appearing behind the Dialog

![image](https://github.com/zakodium-oss/react-science/assets/35760236/41f8c80d-ca93-4d16-88f7-57f86fe594b5)

wadjih-bencheikh18:
@hamed-musallam @stropitek 
This happens because of `react-science` [Portal](https://github.com/zakodium-oss/react-science/blob/main/src/components/root-layout/Portal.tsx ) blueprint js portal are diffrent

So we need to refactor any `react-science` component that uses it.

hamed-musallam:
I believe the color picker should use the Portal from Blueprintjs. As I can see, we have fully replaced the Modal with Dialog.

PR #641, opened 2024-01-30
638 clicking outside to close the color picker is closing the modal

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/638


stropitek:
Can you configure the `Popover` so that it does not display an arrow (blueprintjs docs show that this is possible).

The arrow and the border don't really play nice together:
CleanShot 2024-01-30 at 14 05 09

wadjih-bencheikh18:
> Can you configure the `Popover` so that it does not display an arrow (blueprintjs docs show that this is possible).

Improved

stropitek (approved):

PR #644, opened 2024-02-02
StackedPanel component

### LINKED ISSUE #643: StackedPanel component
API suggestion (if this one does not work let me know)

```jsx
<StackedPanel>
  {({ openConfirmPanel, closePanel }) => (
    <Toolbar>
      <Toolbar.Item
        onClick={openConfirmPanel({
          title: 'title',
          content: <div>Panel content</div>,
        })}
      />
    </Toolbar>
  )}
</StackedPanel>
```

calling `openConfirmPanel` opens a new panel with the close and confirm buttons in the header, places `title` on the left in the header, and content below the header. It's possible to close the panel within the content by passing `closePanel`.

Bonus: think about wether this API can be used to stack more than 1 panel on top of the main content. The idea being that we can render in the `content` property another `StackedPanel`, it will slide over the entire panel again, not just the "content" part of it. I think this requires to add a react context so that it is the same stack of panels (as in blueprint's component's `stack` property) that is edited.

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/643
ref : https://github.com/zakodium-oss/react-science/issues/616
ref : https://github.com/zakodium-oss/react-science/pull/633


wadjih-bencheikh18:
I's useless to have the closePanel function for the initalPanel because it works like that:

```
 /**
     * Call this method to programatically close this panel. If this is the only
     * panel on the stack then this method will do nothing.
     *
     * Remember that the panel header always contains a "back" button that
     * closes this panel on click (unless there is only one panel on the stack).
     */
```

PR #645, opened 2024-02-03
Close select on escape key

### LINKED ISSUE #622: Select keyboard navigation does not work with submenus
It kind of works - sub items can be selected only with the keyboard - but the submenus don't open when a subitem is active.

https://react-science.pages.dev/stories/?path=/story/forms-select--categories-nested

### LINKED ISSUE #632: Select does not close on escape key
This is especially confusing when used in modals because then the modal closes.

In the examples of the documentation of bluprintjs, `Esc` closes the select list.

https://blueprintjs.com/docs/#select/select-component

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/632

closes : https://github.com/zakodium-oss/react-science/issues/622


wadjih-bencheikh18:
@stropitek Fixed, Could you please recheck categories-nested story.
I found that it wasn't practical so I tried to improve it.

stropitek:
Ref: https://github.com/zakodium-oss/react-science/issues/622

Did you base your resolution on https://github.com/palantir/blueprint/pull/4244#issuecomment-850554358?

wadjih-bencheikh18:
> Did you base your resolution on [palantir/blueprint#4244 (comment)](https://github.com/palantir/blueprint/pull/4244#issuecomment-850554358)?

This solution also have the same problem "Keyboard navigation is broken".

wadjih-bencheikh18:
> Ref: #622

Fixed, Could you check please?

stropitek:
Mouse interaction is broken

https://github.com/zakodium-oss/react-science/assets/4118690/19dc747a-b690-4ed9-b7ca-8a3d2228c89c

I think this component library is a great source of inspiration for how accessible components should work:
https://react-spectrum.adobe.com/react-spectrum/Menu.html#dynamic

If possible we should try to mimic their behavior

stropitek:
You can check what radix-ui has in terms of nested menus.

If blueprintjs is too broken maybe it's possible to mix the accessibility of radix-ui with the presentation of blueprintjs

wadjih-bencheikh18:
I fixed what you mentioned.
Otherwise if you prefer other scenario please just describe it.

wadjih-bencheikh18:
> If blueprintjs is too broken

I think blueprintjs components are good but the examples in the documentation are not enough

stropitek:
> I fixed what you mentioned.
> Otherwise if you prefer other scenario please describe it.

It is good enough for now. Anyway I think nested menu aren't really a thing we should use in a select component. It makes more sense in a dropdown menu.

stories/components/select.stories.tsx
```diff
@@ -114,6 +146,15 @@ export function OnlyOptions() {
         filterable={false}
         itemsEqual="label"
         popoverTargetProps={{ style: { display: 'inline-block' } }}
+        popoverProps={{
```
stropitek, line 149: Instead of repeating this over and over, please make a hook `useSelect` which contains the state and returns the `items`, `onItemSelect`, default `itemRenderer` and `popoverProps` to be spread into the `Select` component.

If there are any stories that use a filterable select, check if the hook should be adapted to returns something different based on this parameter (and then also return the `filterable` prop in the hook). Check for any other prop that might be useful to return in the hook based on existing stories.

The hook can be exported by react-science to be used by the consumer of the library.

wadjih-bencheikh18: implemented

stropitek (approved):

src/components/hooks/useState.tsx
stropitek, line 1: Can you just change the name of this file to `useSelect`?

wadjih-bencheikh18: Sorry it was a typo

Issue #607, opened 2023-12-14
Use 14px as base font size


stropitek:
Let's make this change so that our base font is compatible with blueprintjs's font size for regular-sized text.

Blueprint sets a base font of 14px in components, without using relative sizes.

The root layout component currently sets the base font to 12px. It causes some issues. For example, displaying a Select in a modal makes the list appear in a portal that inherits this base size and displays it smaller than it normally would.

stropitek:
TODO:
- Change base font size to 14px
- Review all `fontSize` in the application
  - If expressed in `em` unit will need to be updated to stay the same font size
  - If expressed in pixel, could change to `em` if possible
- Carefully compare app and story between main branch and working branch to spot inconsistencies

PR #652, opened 2024-02-09
Use 14px as base font size

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/607


stropitek:
> The 2 parts highlighted in red were not there before and are added by this PR.

I am making this conclusion based on 
CleanShot 2024-02-12 at 14 13 31

wadjih-bencheikh18:
> It seems like it changes what is displayed in the IR plot.

Yes, I have to fix the tests
Making the default font bigger change the display so the zoom result will change for sure.

wadjih-bencheikh18:
Tests updated

stropitek:
> Making the default font bigger change the display so the zoom result will change for sure.

Yes but a change it font shouldn't make the title of the plot appear / disappear
This hints to a potential bug in the implementation

I can merge since the title is now displayed as it should but maybe worth investigating

Issue #649, opened 2024-02-09
Add "count" (badge) to toolbar item


stropitek:
It adds a small colored "pill" on the corner of the button with the count displayed.

We expect an integer to be passed and I think you can adapt the font size based on if there is 1, 2, or more digits in the number.

Example in NMRium:
![CleanShot 2024-02-09 at 15 42 54](https://github.com/zakodium-oss/react-science/assets/4118690/fbe9e41d-cfe8-4f15-9918-ca8feab66f37)

API suggestion:

```jsx
// text color is always white
<Toolbar.Item count={3} countBackgroundColor="red" />
```

wadjih-bencheikh18:
I like this style, WDYT?

![image](https://github.com/zakodium-oss/react-science/assets/56769876/3c46fc34-1ef2-4bee-9372-e506d0280165)
And on hover
![image](https://github.com/zakodium-oss/react-science/assets/56769876/4a139cda-0a3b-4253-adab-f7add7734f3c)

I used Tag component https://blueprintjs.com/docs/#core/components/tag

stropitek:
SGTM, see PR comments

PR #654, opened 2024-02-10
Add tag to button component

wadjih-bencheikh18:
Closes : https://github.com/zakodium-oss/react-science/issues/649


src/components/button/Button.tsx
```diff
@@ -27,9 +32,33 @@ export function Button(props: ButtonProps) {
...
+                position: 'absolute',
+                top: 0,
+                left: 0,
+                cursor: 'default',
+                fontSize: '0.875em',
+                padding: 2,
+                minHeight: 15,
+                minWidth: 15,
+                lineHeight: '1em',
+                flexDirection: 'column-reverse',
```
stropitek, line 48: Is this really useful?

stropitek (commented):
- Allow `tagProps` to be passed to toolbar component as well. If the `style` prop is passed to `tagProps`, it should probably not override things like "absolute" which are in the default `style`.
- In button and toolbar stories, do not show example with tag in control. Show tag in specific story.
- Stories with longer tags than 1 chars. Up to 3 chars for the toolbar story to see how it looks like in that case.

src/components/button/Button.tsx
wadjih-bencheikh18: Yes to make it centred

stropitek: `align-text: center;` is better then (mainly because more explicit about why the property is there)

wadjih-bencheikh18: I used `flexDirection: 'column'` because the default was `flexDirection: 'row'`, and nothing make it centered

stropitek: It works but isn't it clearer to just use the `align-text` property in order to align (center) the text?

stropitek: Using the dev tools it seems to work as well

stropitek (approved):

stropitek:
@wadjih-bencheikh18 There is a conflict left to resolve

wadjih-bencheikh18:
@stropitek conflict resolved

Issue #651, opened 2024-02-09
Make toolbar measurements toolbar a little nicer


stropitek:
![CleanShot 2024-02-09 at 10 47 10](https://github.com/zakodium-oss/react-science/assets/4118690/a98ecd57-788c-4992-b4c4-02ad4fdc75aa)

The layout currently isn't very nice.

- Move it above the tabs, like in NMRium. Open to other suggestions.
- Find icons for select / unselect all
- Highlight delete button in red

Issue #650, opened 2024-02-09
Show example of stacked panel in demo app


stropitek:
![CleanShot 2024-02-09 at 10 44 36](https://github.com/zakodium-oss/react-science/assets/4118690/bb5eb63b-9f1e-42f9-8542-8189927e574f)

Instead of having a dedicated panel with the measurements info, add a clickable icon to each measurement in the list that opens a stacked panel.

PR #655, opened 2024-02-10
Use stacked panel for measurement info panel

wadjih-bencheikh18:
Closes : https://github.com/zakodium-oss/react-science/issues/650
Closes : https://github.com/zakodium-oss/react-science/issues/651

### DIFF
--- a/src/app/panels/measurements/MeasurementCheckbox.tsx
+++ b/src/app/panels/measurements/MeasurementCheckbox.tsx
@@ -14,6 +14,7 @@ const InputMeasurementCheckBox = styled.input`
   width: 1rem;
   height: 1rem;
   border-width: 1px;
+  margin-left: 10px;
 `;
 
 export function MeasurementCheckbox(props: MeasurementCheckboxProps) {
--- a/src/app/panels/measurements/MeasurementCheckbox.tsx
+++ b/src/app/panels/measurements/MeasurementCheckbox.tsx
@@ -14,7 +14,6 @@ const InputMeasurementCheckBox = styled.input`
   width: 1rem;
   height: 1rem;
   border-width: 1px;
-  margin-left: 10px;
 `;
 
 export function MeasurementCheckbox(props: MeasurementCheckboxProps) {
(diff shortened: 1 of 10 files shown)


wadjih-bencheikh18:
I got insperation from weight and biases platform
image

stropitek:
The new layout with grouped action in the table header is great 👍🏻 

I noticed that the issue with the direction of the animation is still there (so not a storybook issue afterall?). I'm creating another issue for that.

CleanShot 2024-02-12 at 11 38 44

In this PR, can you make this button a dropdown with the choices:
- View metadata - opens the measurements info panel (stacked). It should not have the confirm / close choices, only close. Choose an appropriate icon ("info" icon)
- Configure - opens the measurements configuration panel (stacked). Should also only have close choice, since editing the color of the spectrum has an immediate effect. Don't replace the measurement accordion panel yet, just add it as a stacked panel in the measurements.

wadjih-bencheikh18:
> I noticed that the issue with the direction of the animation is still there (so not a storybook issue afterall?). I'm creating another issue for that.

Not the case for me, which navigator are you using?

targos:
I can reproduce with Chrome and Firefox.

wadjih-bencheikh18:
> In this PR, can you make this button a dropdown with the choices:

Implemented.

stropitek:
Last thing: can the delete button remove selected instead of all.

Remove all should be select all then delete selected

wadjih-bencheikh18:
> I found out that in local is diffrent then here.

Do you have any interpretation of that?
it works perfectly only in the developement version

stropitek:
> Do you have any interpretation of that?
> it works perfectly only in the developement version

It's probably related to strict mode since it is enabled in the app and disabled in the storybook (and always disabled in prod).

So without strict mode it seems to not work.

Their documentation does not seem to be affected by that.
Maybe try to start see if you are able to have a working example based on their documentation. And then from that see how you need to change it to reproduce the bug.

stropitek:
The most notable affect of using strict mode is to re-run effects in developments.

https://react.dev/reference/react/StrictMode

stropitek (approved):

Issue #660, opened 2024-02-12
Measurement plot story broken


stropitek:
https://react-science.pages.dev/stories/?path=/story/layout-measurementplot--control

![CleanShot 2024-02-12 at 14 06 22](https://github.com/zakodium-oss/react-science/assets/4118690/04ae52ed-86d7-48ed-99ad-38aa5baba90a)

PR #661, opened 2024-02-12
Fix measurement-plot story

wadjih-bencheikh18:
Closes : https://github.com/zakodium-oss/react-science/issues/660


stropitek:
@wadjih-bencheikh18 commits that only touch the storybook should be using the `docs` commit prefix

Issue #662, opened 2024-02-12
Use blueprintjs Radio in RadioGroup


stropitek:
Keep the `RadioGroup` component

Replace "classic" implementation with an implementation that uses blueprintjs `Radio`

PR #664, opened 2024-02-12
Use blueprintjs Radio in RadioGroup

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/662


stropitek (commented):
You can remove our legacy "classic" implementation of the radio button

You can remove the "type" prop from our component and always render the button variant.

You remove the story that renders both variants side by side and replace with 2 control components, one for each component.

PS: there is an eslint error

stropitek (approved):

PR #666, opened 2024-02-16
Add hover option for popover toolbar trigger

### LINKED ISSUE #658: Add story with hover popover toolbar items
Implement new story where the trigger is not click but hover

moonayyur:
closes #658

### DIFF
--- a/src/components/toolbar/Toolbar.tsx
+++ b/src/components/toolbar/Toolbar.tsx
@@ -28,6 +28,12 @@ import {
   useToolbarContext,
 } from './toolbarContext';
 
+export type PopoverInteractionType =
+  | 'click'
+  | 'hover'
+  | 'click-target'
+  | 'hover-target';
+
 interface ToolbarBaseProps {
   intent?: Intent;
   disabled?: boolean;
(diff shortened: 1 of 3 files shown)


stropitek (approved):

src/components/toolbar/Toolbar.tsx
```diff
@@ -56,6 +62,7 @@ export interface ToolbarItemProps extends ToolbarBaseProps {
 
 export interface ToolbarPopoverItemProps extends PopoverProps {
   itemProps: ToolbarItemProps;
+  popoverInteraction?: PopoverInteractionType;
```
stropitek, line 65: You did not need to invent a new prop to add the behaviour to the item. Popover props are already spread into the `Popover` component so you could have used the original prop for this.

However, it makes sense to pass this option to the `Toolbar` component via a `popoverInteractionKind` prop to avoid having to pass this to every item.

moonayyur: I added the `popoverInteractionKind` prop to the toolbar context

stropitek:
Can you shorten the the duration of the popover transition so that menus don't overlap?

Don't know if it should be 0 or just something smaller.

stropitek (approved):

Issue #674, opened 2024-02-23
The "DropZone" component should have an option to prevent the propagation of drop events.


hamed-musallam:
https://react-dropzone.js.org/#section-event-propagation
https://github.com/zakodium-oss/react-science/blob/c4628bd14c2f0258436cbfa4f3a24b7a48f44a8a/src/components/drop-zone/DropZone.tsx#L107-L111

We need to have this option to stop propagation from the child to the parent, where we have a similar case in NMRium

https://github.com/zakodium-oss/react-science/assets/35760236/99cc3161-d1cb-4b19-9c24-a4ef4e846438

PR #675, opened 2024-02-26
feat!: extend drop zone component properties 

hamed-musallam:


stropitek (commented):
The commit message should be `feat` with an additional message indicating that there is a breaking change for the `fileValidator` prop.

Otherwise the changes LGTM.

Also it would be great if this PR could improve the stories a little bit. For example handling the validation error.

hamed-musallam:
@stropitek 

could you merge this PR because I need this hotfix to fix an issue in the NMRium? https://github.com/zakodium-oss/react-science/issues/674

PR #677, opened 2024-03-07
fix: update dependencies

lpatiny:


targos:
```
Error: Cannot find module @rollup/rollup-linux-x64-gnu. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
```

Have you removed `node_modules` and `package-lock.json` before doing `npm install`?

PR #681, opened 2024-03-13
feat: improve use select hook

hamed-musallam:
Improve the useSelect hook :

- It accepts a `defaultSelectedItem` property.
- You can specify which object serves as the label in two ways:
    - Using the `itemTextKey` property, which accepts only keys where the value type is a string.
    - Using the `getItemText` function, which takes an item as a parameter and returns a string.

### DIFF
--- a/src/components/hooks/useSelect.tsx
+++ b/src/components/hooks/useSelect.tsx
@@ -10,12 +10,12 @@ type FilterType<SourceType, Type> = Pick<
 >;
 
 interface BaseOptions<T> {
-  itemLabelKey: keyof FilterType<T, string>;
+  itemTextKey: keyof FilterType<T, string>;
   defaultSelectedItem?: T;
 }
 
 interface RenderOptions<T> {
-  itemLabelMap: (item: T) => string;
+  getItemText: (item: T) => string;
   defaultSelectedItem?: T;
 }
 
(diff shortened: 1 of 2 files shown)


hamed-musallam:
@stropitek 

could you take a look at this PR, Now the `useSelect` hook 

- It accepts a `defaultSelectedItem` property.
- You can specify which object serves as the label in two ways:
    - Using the `itemLabelKey` property, which accepts only keys where the value type is a string.
    - Using the `itemLabelMap` function, which takes an item as a parameter and returns a string.

src/components/hooks/useSelect.tsx
```diff
@@ -2,9 +2,50 @@ import { MenuItem } from '@blueprintjs/core';
...
+  }[keyof SourceType]
+>;
+
+interface BaseOptions<T> {
+  itemLabelKey: keyof FilterType<T, string>;
+  defaultSelectedItem?: T;
+}
+
+interface RenderOptions<T> {
+  itemLabelMap: (item: T) => string;
```
stropitek, line 18: Can you rename this to `getText`? (because it maps to the `text` prop in blueprintjs)

src/components/hooks/useSelect.tsx
```diff
@@ -2,9 +2,50 @@ import { MenuItem } from '@blueprintjs/core';
...
-  const itemRenderer = getItemRenderer(value);
+type FilterType<SourceType, Type> = Pick<
+  SourceType,
+  {
+    [K in keyof SourceType]: SourceType[K] extends Type ? K : never;
+  }[keyof SourceType]
+>;
+
+interface BaseOptions<T> {
+  itemLabelKey: keyof FilterType<T, string>;
```
stropitek, line 13: Rename to `itemTextKey`

src/components/hooks/useSelect.tsx
stropitek: Or `getItemText`

stropitek (approved):

Issue #679, opened 2024-03-08
Reduce size of toolbar in the panels


lpatiny:
The icons in the header toolbar of panel should be slightly smaller that the one in the normal toolbar

![image](https://github.com/zakodium-oss/react-science/assets/1484241/1f2c8ed3-70ee-4bb5-8d37-4ed814e930c8)

wadjih-bencheikh18:
https://github.com/zakodium-oss/react-science/pull/693

Do we close this issue?

PR #693, opened 2024-03-18
Reduce size of toolbar in the panels

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/679


stropitek:
I'm not in favour of this. The icons look blurry.

The icons are made to be displayed 16x16 and there is not much we can do about it because smaller versions of the icons do not exist.

@lpatiny

targos:
16x16 seems to be a standard size for small icons, we should not try to make it smaller or people will have trouble to see them.

wadjih-bencheikh18:
@targos Do I close this PR?

targos:
I think so, yes

Issue #691, opened 2024-03-16
Closing a dialog opens trigger tooltip


targos:
Reproduce:
- Go to https://react-science.pages.dev/pages/demo
- Click on the about button (top-left)
- Press escape
- "About Logo" tooltip appears.

I think it's a case of https://blueprintjs.com/docs/#core/components/tooltip.combining-with-popover and we should decide to either set `shouldReturnFocusOnClose={false}` on the relevant dialogs, or `openOnTargetFocus={false}` on the `Toolbar.Item`.

/cc @stropitek

stropitek:
setting `shouldReturnFocusOnClose={false}` is a better option (for UX, probably not for DX because cannot enforce it on every dialog) IMO

targos:
This affects a bunch of dialogs on NMRium that are opened from a toolbar button.

PR #695, opened 2024-03-22
refactor: disable return-focus-on-close on Dialog

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/691

### DIFF
--- a/src/components/dialog/ConfirmDialog.tsx
+++ b/src/components/dialog/ConfirmDialog.tsx
@@ -31,7 +31,6 @@ export function ConfirmDialog(props: ConfirmDialogProps) {
 
   return (
     <Dialog
-      shouldReturnFocusOnClose={false}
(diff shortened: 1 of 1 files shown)


src/components/dialog/ConfirmDialog.tsx
```diff
@@ -31,6 +31,7 @@ export function ConfirmDialog(props: ConfirmDialogProps) {
 
   return (
     <Dialog
+      shouldReturnFocusOnClose={false}
```
targos, line 34: Not sure about this one. I think it should be up to the application to pass this prop, and usually confirm dialogs shouldn't come from elements that have a tooltip.

wadjih-bencheikh18: Updated thank you

PR #699, opened 2024-03-27
feat!: improve toolbar tip

hamed-musallam:
Screenshot 2024-03-27 at 15 51 18

### DIFF
--- a/src/components/toolbar/Toolbar.tsx
+++ b/src/components/toolbar/Toolbar.tsx
@@ -52,7 +52,9 @@ export interface ToolbarProps extends ToolbarBaseProps {
   popoverInteractionKind?: PopoverInteractionType;
 }
 
-export interface ToolbarItemProps extends ToolbarBaseProps {
+export interface ToolbarItemProps
+  extends ToolbarBaseProps,
+    Pick<TooltipProps, 'minimal'> {
   id?: string;
   hoverContent: TooltipProps['content'];
   icon: IconName | JSX.Element;
@@ -63,7 +65,6 @@ export interface ToolbarItemProps extends ToolbarBaseProps {
   isPopover?: boolean;
   tag?: ReactNode;
   tagProps?: Omit<TagProps, 'children'>;
-  noArrow?: boolean;
 }
 
 export interface ToolbarPopoverItemProps extends PopoverProps {
@@ -148,7 +149,7 @@ Toolbar.Item = function ToolbarItem(props: ToolbarItemProps) {
     disabled: itemDisabled,
     noHoverContent = false,
     isPopover,
-    noArrow = false,
+    minimal,
     ...other
   } = props;
 
@@ -219,7 +220,7 @@ Toolbar.Item = function ToolbarItem(props: ToolbarItemProps) {
         noHoverContent
           ? undefined
           : {
-              modifiers: { arrow: { enabled: !noArrow } },
+              minimal,
               content: hoverContent,
               placement: vertical ? 'right' : 'bottom',
               intent,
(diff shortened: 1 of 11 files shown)


src/components/toolbar/Toolbar.tsx
```diff
@@ -62,6 +65,7 @@ export interface ToolbarItemProps extends ToolbarBaseProps {
   isPopover?: boolean;
   tag?: ReactNode;
   tagProps?: Omit<TagProps, 'children'>;
+  noArrow?: boolean;
```
stropitek, line 68: Have you tried to use the `minimal` prop on the `Toolbar` component? I think it does what you need.

src/components/toolbar/Toolbar.tsx
```diff
@@ -146,6 +150,8 @@ Toolbar.Item = function ToolbarItem(props: ToolbarItemProps) {
     disabled: itemDisabled,
     noTooltip = false,
     isPopover,
+    noArrow = false,
+    interactionKind,
```
stropitek, line 154: The interaction kind should be set on the Toolbar (you already can via `popoverInteractionKind`)

I don't think having different interaction on items of the same toolbar is good UX.

src/components/toolbar/Toolbar.tsx
hamed-musallam: there is no `minimal` property over the `Toolbar` component, it is just over the ToolbarPopoverItem and the tooltip not working there all

src/components/toolbar/Toolbar.tsx
stropitek: Have you tried using `Toolbar.PopoverItem`?

hamed-musallam: yes, it works well but the tooltip not working
image
image

hamed-musallam: > Toolbar.PopoverItem

Sorry for the confusion. I initially thought you were referring to `ToolbarPopoverItem`, but actually, I'm using `ToolbarPopoverItem` and `Toolbar.Item` components.

stropitek: `ToolbarPopoverItem` == `Toolbar.PopoverItem`

My suggestion is that you use `Toolbar.PopoverItem` component to render some custom content instead of the simple tooltip of the `Toolbar.Item` component. If something is not working as expected with this component then I can look into it if you provide me with some more detailed explanations of the issue.

On `Toolbar.Item`, the `title` prop should remain `string`. 

You will probably want to add the ability to set or not the little "chevron" on the toolbar button (they make sense from menus, not for the type of popover content you created in your story).

src/components/toolbar/Toolbar.tsx
hamed-musallam: It is done, we remove the `noArrow ` property and property and replaced it with `minimal ` over the `Toolbar.Item` component

hamed-musallam:
@stropitek 

Is it okay for you to have hoverContent=""?

https://github.com/zakodium-oss/react-science/blob/bdbe1ebb0fbe177d2967891f4cc151540dbf615b/src/components/toolbar/Toolbar.tsx#L264

stories/components/toolbar.stories.tsx
```diff
@@ -401,3 +401,149 @@ MixedItems.argTypes = {
...
+
+export function CustomTipContent() {
+  return (
+    <div
+      style={{
+        display: 'flex',
+        height: 200,
+      }}
+    >
+      <Toolbar vertical popoverInteractionKind="hover">
```
stropitek, line 461: the `popoverInteractionKind` only has an effect on `PopoverItem`

stropitek (approved):

Issue #449, opened 2023-02-24
New panel for logs


lpatiny:
We have a new library, fifo-logger, that allows to manage logs and we plan to use in various projects.

We should be able to display the resulting logs in a table. Here is an example that can be used: 

Property could be named `logs`

We should display:
- Label: levelLabel
- Message: message
- An icon to display the stack trace if it is an error (presence of error property)
- An icon to display the meta information if there are any properties in meta

We should also be able to filter the list by 'level' based on the following list
- 10: trace
- 20: debug
- 30: info
- 40: warn
- 50: error
- 60: fatal

Light background color for the line could highlight the levels:
- pink for error and fatal
- lightyellow for warn
- lightgreen for info, debug and trace

```
    [
      {
        "level": 30,
        "time": 1677268099762,
        "message": "This is an info",
        "meta": {},
        "levelLabel": "info"
      },
      {
        "level": 40,
        "time": 1677268099763,
        "message": "This is a warning with object",
        "meta": {},
        "levelLabel": "warn"
      },
      {
        "level": 40,
        "time": 1677268099763,
        "message": "This is a warning",
        "meta": {
          "a": 1,
          "b": 2,
          "c": "Hello"
        },
        "levelLabel": "warn"
      },
      {
        "level": 50,
        "time": 1677268099763,
        "context": "c71f3fbf-dabf-4b16-b764-9e69a066f151",
        "message": "This is an error in a child",
        "meta": {},
        "levelLabel": "error"
      },
      {
        "level": 60,
        "time": 1677268099763,
        "context": "c71f3fbf-dabf-4b16-b764-9e69a066f151",
        "message": "Fatal error",
        "error": {
          "type": "Error",
          "message": "Fatal error",
[...]

PR #709, opened 2024-04-05
feat: add logger components

stropitek:
Closes: https://github.com/zakodium-oss/react-science/issues/449


stropitek:
Usage:

```jsx
import { FifoLogger } from 'fifo-logger';
const fifoLogger = new FifoLogger();

function App() {
  return (
    <FifoLoggerProvider logger={fifoLogger}>
      // ...
      <FifoLoggerToolbarItem />
    </FifoLoggerProvider>
  );
}
```

To add logs:

```jsx
function ComponentAddingLogs() {
  const logger = useFifoLogger();

  // In callback / useEffect
  logger.info('info');
}
```

To get render something when the logs are updated:

```jsx
function ComponentThatNeedsLogs() {
  const logs = useFifoLogs();
}
```

stropitek:
@lpatiny This PR is awaiting your review

@moonayyur can you have a look as well? Because you designed a similar component in pixelium and the goal is to replace it with this one.

lpatiny (approved):
Functionality LGTM but there is no storybook so it may be difficult to find this component,

moonayyur (approved):

package.json
```diff
@@ -60,6 +60,7 @@
   "peerDependencies": {
     "@blueprintjs/core": "^5.3.2",
     "@blueprintjs/icons": "^5.1.7",
+    "fifo-logger": "^1.0.0",
```
targos, line 63: I'm not sure why it should be a peerDependency?

targos: Because only the types are used.

PR #715, opened 2024-04-08
feat: tooltip help content

hamed-musallam:


targos:
I don't understand the name. Why "Template" ?

hamed-musallam:
> I don't understand the name. Why "Template" ?

It is not easy to get a good name for the component, I use `Template` because the component is a new layout for the tooltip, Do you have any alternative suggestions?

src/components/toolbar/TooltipTemplate.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+const styles: Record<
+  | 'titleContainer'
+  | 'title'
+  | 'description'
+  | 'shortcutContainer'
+  | 'shortcutItem',
+  CSSProperties
+> &
+  Record<'subTitleListItem' | 'subTitleList', SerializedStyles> = {
```
stropitek, line 16: I wrote some guidelines on how to style things in react-science. Please have a look at https://github.com/zakodium-oss/react-science/wiki/Contributor-guidelines#styling-components

Good example in react-science: https://github.com/zakodium-oss/react-science/blob/main/src/components/drop-zone/DropZone.tsx

The guidelines are and we can update them if more usage patterns are needed.

stropitek (approved):

stories/components/toolbar.stories.tsx
```diff
@@ -473,77 +469,25 @@ export function CustomTipContent() {
...
-
-      <p style={styles.description}>
-        {description}{' '}
-        <a target="_blank" href="">
-          Learn more
-        </a>{' '}
-      </p>
+    <div>
+      <TooltipHelpContent
+        style={{ backgroundColor: 'rgb(62, 62, 62)' }}
```
stropitek, line 476: Can you remove this?

I don't think we need to show that we can change the background color and it's not obvious that we do that in the story.

hamed-musallam: Without the background, the tooltip content won't be visible here https://tooltip-template.react-science.pages.dev/stories/?path=/story/components-toolbar--tooltip-content

stropitek: I see, can you setup controls on https://tooltip-template.react-science.pages.dev/stories/?path=/story/components-toolbar--custom-tooltip-content so that we can see how changing the intent affects the tooltip content?

targos: I added a Tooltip wrapper so we can inherit the intent and see it with the expected style.

Issue #721, opened 2024-04-19
Invalid prop passed to table row in info panel


targos:
![CleanShot 2024-04-19 at 09 53 56](https://github.com/zakodium-oss/react-science/assets/2352663/74da953f-bc4b-4663-a1bf-7166bd049ebf)

PR #723, opened 2024-05-02
Fix InfoPanel Table Row

wadjih-bencheikh18:
closes https://github.com/zakodium-oss/react-science/issues/721

### DIFF
--- a/src/app/panels/SignalProcessingPanel.tsx
+++ b/src/app/panels/SignalProcessingPanel.tsx
@@ -47,7 +47,7 @@ export function SignalProcessingPanel(props: SignalProcessingPanelProps) {
           <ValueRenderers.Header value="Options" />
         </Table.Header>
         {filters.map(({ name, options }, i) => (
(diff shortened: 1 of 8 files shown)


wadjih-bencheikh18:
What's the difference between `bordered` and `hasBorder` in Table props?
Maybe we can remove the context hook
https://github.com/zakodium-oss/react-science/blob/287f1c3eff668e687ca5a19b04d23fd42185b6bf/src/components/table/Table.tsx#L33-L37

stropitek:
I don't know. Are both used by the implementation?

wadjih-bencheikh18:
> I don't know. Are both used by the implementation?

`bordered` is used directly by `HtmlTable` of blueprintjs
and `hasBorder` is the one used in the context
the two add border to the table so technically the two props do the same thing

stropitek:
I think you can remove `hasBorder` all together and only keep the blueprintjs prop.

stropitek (approved):

Issue #710, opened 2024-04-05
Do not scroll behind the search box in info panel


lpatiny:
In the info panel I would prefer that the table does not scroll behind the search box and that the search box position / size is fixed like in:

![2024-04-05 15 09 43](https://github.com/zakodium-oss/react-science/assets/1484241/e6bf6d72-1c68-4854-b6a4-51ace6bb6ead)

wadjih-bencheikh18:
which option do you prefer
1
![image](https://github.com/zakodium-oss/react-science/assets/56769876/63977fdc-e767-442c-96a7-53f7396d3a82)

2
![image](https://github.com/zakodium-oss/react-science/assets/56769876/50e701ae-35cb-459d-b4c2-599d60614a37)

lpatiny:
I prefer you repeat the headers (so indeed 2)

PR #727, opened 2024-05-22
feat: improve InfoPanel

wadjih-bencheikh18:
closes https://github.com/zakodium-oss/react-science/issues/710


lpatiny (approved):
I tried the branch and indeed the scroll bar is now at the correct place.

image

stropitek:
@lpatiny is this ok? Seems not great that the panel does not reveal its content when opened, because it is not at the right scroll position.

https://github.com/zakodium-oss/react-science/assets/4118690/c4bad4e5-f06f-4d8e-9e55-fc5b729d29b9

targos:
I think there should be I lay one sticky header and they should not accumulate

lpatiny:
Indeed I didn't test with 2 sections and it is very confusing currently. I was not thinking that the title of the section were not scrolling but only that it does not scroll behind the search bar. On the other hand seeing in which section you are is an added value as well.

I agree with Michael that we should not accumulate the sticky header because it does not bring anything and it yields to the issue Daniel pointed out.

stropitek (approved):

PR #729, opened 2024-05-24
feat: add sticky table header option

### LINKED ISSUE #728: Add option to make table header sticky

moonayyur:
closes #728

### DIFF
--- a/src/components/table/Table.tsx
+++ b/src/components/table/Table.tsx
@@ -34,8 +34,16 @@ export interface TableProps extends React.HTMLAttributes<HTMLTableElement> {
   compact?: boolean;
   interactive?: boolean;
   striped?: boolean;
+  stickyHeader?: boolean;
   color?: string;
 }
+
+const headerStyle: React.CSSProperties = {
+  position: 'sticky',
+  top: 0,
+  zIndex: 10,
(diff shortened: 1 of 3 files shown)


stropitek (commented):
Why didn't you put the sticky option in the `Header` component?


CleanShot 2024-05-27 at 10 33 04

We shouldn't be able to see the table rows which are above the header.

moonayyur:
> Why didn't you put the sticky option in the `Header` component?

Should I use `cloneElement` to pass the prop to `Header` then ?

stropitek:
> Should I use `cloneElement` to pass the prop to `Header` then ?

Can you explain? I don't think you need to do that in order to move the sticky prop from the `Table` component to the `TableHeader` component.

src/components/table/Table.tsx
```diff
@@ -34,6 +34,7 @@ export interface TableProps extends React.HTMLAttributes<HTMLTableElement> {
   compact?: boolean;
   interactive?: boolean;
   striped?: boolean;
+  stickyHeader?: boolean;
```
stropitek, line 37: This should only be in the props of the `TableHeader` component and be called `sticky`.

stropitek (approved):

PR #730, opened 2024-05-27
fix: disable button tooltips with empty content

hamed-musallam:
We have a warning in the console from Blueprint.js when the button component is used with empty tooltip content. To prevent this, we should disable the tooltip if its content is empty, which makes sense since an empty tooltip should be avoided.

I believe Blueprint.js should disable the tooltip by default if the content is empty.

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -18,7 +18,7 @@ type BlueprintProps = {
     BlueprintAnchorButtonProps[key];
 };
 export type ButtonProps = BlueprintProps & {
-  tooltipProps?: Omit<TooltipProps, 'children'>;
+  tooltipProps?: Partial<Omit<TooltipProps, 'children'>>;
   tag?: ReactNode;
   tagProps?: Omit<TagProps, 'children'>;
 };
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -24,20 +24,20 @@ export type ButtonProps = BlueprintProps & {
 };
 
 export function Button(props: ButtonProps) {
-  const { tooltipProps, children, tag, tagProps, ...buttonProps } = props;
+  const { tooltipProps = {}, children, tag, tagProps, ...buttonProps } = props;
   const {
     fill,
     content = '',
-    disabled = false,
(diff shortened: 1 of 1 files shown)


src/components/button/Button.tsx
```diff
@@ -25,14 +25,21 @@ export type ButtonProps = BlueprintProps & {
 
 export function Button(props: ButtonProps) {
   const { tooltipProps, children, tag, tagProps, ...buttonProps } = props;
```
stropitek, line 27:
```suggestion
  const { tooltipProps = {}, children, tag, tagProps, ...buttonProps } = props;
```

src/components/button/Button.tsx
```diff
@@ -25,14 +25,21 @@ export type ButtonProps = BlueprintProps & {
...
-
+  const {
+    fill,
+    content = '',
+    disabled = false,
+    ...otherToolTipProps
+  } = tooltipProps || {};
```
stropitek, line 33:
```suggestion
  } = tooltipProps;
```

src/components/button/Button.tsx
```diff
@@ -25,14 +25,21 @@ export type ButtonProps = BlueprintProps & {
 
 export function Button(props: ButtonProps) {
   const { tooltipProps, children, tag, tagProps, ...buttonProps } = props;
-
+  const {
+    fill,
+    content = '',
+    disabled = false,
```
stropitek, line 31:
```suggestion
    disabled = !tooltipProps.content,
```

src/components/button/Button.tsx
```diff
@@ -25,14 +25,21 @@ export type ButtonProps = BlueprintProps & {
...
+  } = tooltipProps || {};
   const InnerButton = buttonProps.disabled
     ? BlueprintAnchorButton
     : BlueprintButton;
   return (
     <Tooltip
-      fill={tooltipProps?.fill || buttonProps.fill}
-      {...tooltipProps}
+      fill={fill || buttonProps.fill}
+      disabled={!content ? true : disabled}
```
stropitek, line 40:
```suggestion
      disabled={disabled}
```

stropitek (approved):

PR #732, opened 2024-05-28
feat: improve useSelect hook item render

hamed-musallam:
In some cases, we need more than just text in the menu items, we require additional elements, as seen in the workspaces dropdown menu in NMRium. Therefore, I suggest enhancing the useSelect hook to support custom item rendering. I don't want to replace the entire MenuItem; I just need the ability to customize its content.

### DIFF
--- a/src/components/hooks/useSelect.tsx
+++ b/src/components/hooks/useSelect.tsx
@@ -9,16 +9,16 @@ type FilterType<SourceType, Type> = Pick<
   }[keyof SourceType]
 >;
 
-interface ItemRenderOptions<T> {
+interface ItemOptions<T> {
   renderItem?: (item: T) => ReactNode;
   defaultSelectedItem?: T;
 }
 
-interface BaseOptions<T> extends ItemRenderOptions<T> {
+interface BaseOptions<T> extends ItemOptions<T> {
   itemTextKey: keyof FilterType<T, string>;
 }
 
-interface RenderOptions<T> extends ItemRenderOptions<T> {
+interface RenderOptions<T> extends ItemOptions<T> {
   getItemText: (item: T) => string;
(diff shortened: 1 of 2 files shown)


src/components/hooks/useSelect.tsx
```diff
@@ -19,7 +19,9 @@ interface RenderOptions<T> {
   defaultSelectedItem?: T;
 }
 
-type SelectOptions<T> = BaseOptions<T> | RenderOptions<T>;
+type SelectOptions<T> = {
+  renderItem?: (item: T) => ReactNode;
+} & (BaseOptions<T> | RenderOptions<T>);
```
stropitek, line 24: Can you revert this change and instead make `BaseOptions` and `RenderOptions` extend a type that has `renderItem` and `defaultSelectedItem` since those 2 properties are in common?

stropitek (approved):
LGTM modulo my review comment.

src/components/hooks/useSelect.tsx
```diff
@@ -91,16 +94,19 @@ function getItemRenderer<T>(value: T, options: SelectOptions<T>) {
...
     const label = getLabel(item, options);
+    const { renderItem } = options;
+    const isSelected = selectedLabel === label;
+    const { active, disabled } = modifiers;
     return (
       <MenuItem
-        active={modifiers.active}
-        disabled={modifiers.disabled}
-        selected={selectedLabel === label}
+        active={active && isSelected}
```
stropitek, line 102: This breaks the component in a bad way.

We no longer know which item is active when using the keyboard to navigate between items.

stropitek (changes requested):

stropitek (approved):

Issue #469, opened 2023-04-04
Switch components from headlessui or radix to blueprint


targos:
```[tasklist]
### Tasks
- [x] GradiantSelect
- [x] DropdownMenu
- [x] MenuItems
- [x] Dropdown stories
- [x] Remove headlessui dependencies
```

targos:
@stropitek We should review this.

stropitek:
Use radix-ui's `Collapsible` instead of headless-ui's `Disclosure`
Use radix-ui's `Select` instead of headless-ui's `Listbox`

stropitek:
@wadjih-bencheikh18 can you prioritize this?

targos:
The components were switched to radix-ui, not to blueprint, so this issue is not finished.

targos:
Still not finished. I can see radix-ui used in:
- InfoPanel
- RadioGroup
- ButtonRadioItem

PR #734, opened 2024-06-02
469 switch components from headlessui or radix to blueprint

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/469


stropitek (commented):
### Gradient selector

I find it a bit hard to notice the active element on the gradient selector. Can you find something that makes it pop out a little more when active, but still different from the selected option?

### Info panel

- The sections are now collapsed on the first render of the component. I think it was better when they were all open, like before.
- The cursor was `pointer` before, now it's `text`

stropitek:
I did not like changing the opacity because the colours are important and opacity is changing how they actually would look.

I made another implementation, it's not perfect but I think it's good enough for now.

Issue #243, opened 2022-10-13
Create ToolWindow component


targos:
Similar to VSCode with an icon to open it.

stropitek:
Do a POC based on:

https://github.com/cheminfo/nmrium/issues/2782#issuecomment-1846545829

I would suggest that we create a new component `ToolbarPanels`

```jsx
return (
<ToolbarPanels panels={[
    { id: 'panel1', renderPanel: () => <Accordion.Item ... />, icon: 'settings", isOpen: true }
  ]}
/>
```

The component could either be used as a controlled or uncontrolled component.

It would be in uncontrolled mode if another prop is passed (e.g. `initialOpen={['panel1']}`}. This is just a suggestion and the POC should focus on the controlled component.

targos:
I don't see the relation with toolbars (for the name).

stropitek:
> I don't see the relation with toolbars (for the name).

Yes on second thought it's not semantically related.

In Webstorm they are called "Tool windows", in VS code the "Activity bar", I'm okay with the former.

stropitek:
@wadjih-bencheikh18 are you willing to take this on?

We want to do something that looks like a vertical toolbar, but opens panel windows. To give you something more visual, it would behave similarly to webstorm's "Tool windows":

https://github.com/zakodium-oss/react-science/assets/4118690/aa5d9ade-296a-4823-8b07-9ba80761cba9

VSCode also has a similar concept but calls it "Activity bar".

Start by making the simplest possible POC:
- A vertical toolbar, slighly bigger than our existing toolbar (including bigger icons)
- Each tool can be on or off, and you can toggle their state by clicking on the toolbar.
- Tools which are "on" render an accordion with some placeholder content. The accordions can be collapsed as usual.

PR #735, opened 2024-06-12
Implement Panel Toolbar

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/243


wadjih-bencheikh18:
@stropitek this PR is ready to merge

stropitek (commented):
In a real application, the accordion panels will be rendered within a SplitPane, but the toolbar will be rendered some place else in the layout (this part has a fixed size).

So I think the API I proposed is not going to work, because it forces both to be rendered together.

Can you just focus on a simple stateless component that renders the big toolbar buttons and remove the rest of the code?

The component should:
- Have a good name (I'm thinking `PanelToolbar` + `PanelToolbar.Item`, but not convinced)
- Be able to render the items with a prop that makes it look selected / unselected.

wadjih-bencheikh18:
> Can you just focus on a simple stateless component that renders the big toolbar buttons and remove the rest of the code?

Is it worth to create a component for that? isn't it enough to create just story for it. because the implementation is simply div with toolbar items inside.

stropitek:
> Is it worth to create a component for that? isn't it enough to create just story for it. because the implementation is simply div with toolbar items inside.

Indeed if we don't need custom style and just layout with existing components then a story is enough.

src/components/accordion/Accordion.tsx
```diff
@@ -79,6 +79,7 @@ Accordion.Item = function AccordionItem(props: AccordionItemProps) {
 
   return (
     <div
+      key={title}
```
stropitek, line 82:
```suggestion
```

src/components/toolbar/Toolbar.tsx
```diff
@@ -53,21 +50,17 @@ export interface ToolbarProps extends ToolbarBaseProps {
   popoverInteractionKind?: PopoverInteractionType;
 }
 
-export interface ToolbarItemProps extends ToolbarBaseProps {
-  id?: string;
+export interface ToolbarItemProps
+  extends ToolbarBaseProps,
+    Omit<ButtonProps, 'onClick' | 'tooltipProps'> {
```
stropitek, line 55: I would rather `Pick` some props that we know we need, rather than omit.

Each props that we pick should have a corresponding story.

stories/components/panel-toolbar.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+        display: 'flex',
+      }}
+    >
+      <div>
+        <Toolbar vertical>
+          {itemsBlueprintIcons.map((item) => (
+            <Toolbar.Item
+              minimal={false}
+              large
+              fill
```
stropitek, line 47: Those props should be part of the `Toolbar` component. I don't think we should be able to mix different styles of buttons in a toolbar.

When we do we should have stories showing what purpose this mixing serves.

src/components/toolbar/Toolbar.tsx
```diff
@@ -51,17 +53,21 @@ export interface ToolbarProps extends ToolbarBaseProps {
...
+  icon: IconName | ReactElement;
+  active?: boolean;
   onClick?: (
     item: ToolbarItemProps & {
       event: MouseEvent;
     },
   ) => void;
+  className?: string;
   isPopover?: boolean;
+  tag?: ReactNode;
```
stropitek, line 69: Can you use `Pick<ButtonProps, ...>` for all the properties that are forwarded to the `Button` component?

stropitek (changes requested):
The non-minimal version of the toolbar does not seem to be correct.

Screenshot from storybook:
CleanShot 2024-08-19 at 08 54 52

Screenshot from blueprintjs documentation:
CleanShot 2024-08-19 at 08 55 46

src/components/button/Button.tsx
```diff
@@ -41,33 +42,39 @@ export function Button(props: ButtonProps) {
...
-          )}
-          <InnerButton {...targetProps} {...buttonProps} style={{}}>
-            {children}
-          </InnerButton>
-        </div>
+        <InnerButton
+          {...targetProps}
+          {...buttonProps}
+          icon={
+            <div>
```
stropitek, line 49: Can't this be a fragment?

src/components/toolbar/Toolbar.tsx
```diff
@@ -40,33 +38,23 @@ interface ToolbarBaseProps {
...
   tooltip?: TooltipProps['content'];
   tooltipProps?: Omit<TooltipProps, 'content'>;
-  icon: IconName | ReactElement;
-  active?: boolean;
   onClick?: (
     item: ToolbarItemProps & {
       event: MouseEvent;
     },
   ) => void;
-  className?: string;
```
stropitek, line 66: Why have you removed this prop?

stories/components/panel-toolbar.stories.tsx
stropitek, line 1: Can you move this story with the other toolbar stories?

stories/components/toolbar.stories.tsx
```diff
@@ -21,6 +21,8 @@ export default {
     vertical: false,
     intent: 'none',
     disabled: false,
+    fill: false,
+    alignText: 'center',
```
stropitek, line 25: Is `alignText` really a prop of toolbar? What does it do?

stories/components/toolbar.stories.tsx
```diff
@@ -21,6 +21,8 @@ export default {
     vertical: false,
     intent: 'none',
     disabled: false,
+    fill: false,
```
stropitek, line 24: What does it do? When I change it in the story it does not have any effect.

stories/components/toolbar.stories.tsx
```diff
@@ -336,6 +360,7 @@ export function PopoverItems(
               setActive(itemProps.id);
               onClick();
             },
+            tag: (itemProps.id.startsWith('credit-card') && 1) || null,
```
stropitek, line 363:
```suggestion
            tag: itemProps.id.startsWith('credit-card') ? 1 : null,
```

stropitek (commented):
Please review any missing controls (like `minimal` and `vertical`) in [the control story](https://243-create-toolwindow-compon.react-science.pages.dev/stories/?path=/story/components-toolbar--control) and add them to the story.

---

Please write your commit message for this PR. It needs to mention the breaking change for the `Toolbar` component, which now renders differently by default (without `minimal` set to `true`).

wadjih-bencheikh18:
> Please review any missing controls (like `minimal` and `vertical`) in [the control story](https://243-create-toolwindow-compon.react-science.pages.dev/stories/?path=/story/components-toolbar--control) and add them to the story.

Improved

stories/components/toolbar.stories.tsx
```diff
@@ -21,6 +21,9 @@ export default {
     vertical: false,
     intent: 'none',
     disabled: false,
+    minimal: false,
```
stropitek, line 24: Can you make this true since it used to be the default and we actually mostly want to use the minimal version.

stropitek (commented):
Blueprint icons come in 2 sizes: 16px and 20px.

It would make sense that the large version of the component also shows the larger icons, no?

Can you do that so that we can see if it looks better?

wadjih-bencheikh18:
> It would make sense that the large version of the component also shows the larger icons, no?

Yes. Blueprint Button use the small icon in all cases https://blueprintjs.com/docs/#core/components/buttons

wadjih-bencheikh18:
> Can you do that so that we can see if it looks better?

You can take a look.
For the toolbar icons they can be **string** (blueprint native icons) or **component** or **text** 
I tried to check all these cases

stropitek:
OK let's go with the larger icons.

I noticed that when you use the vertical version with the `fill` option to `true`, it fills the horizontal space instead of the vertical space. Can you check if it can be changed to be like https://blueprintjs.com/docs/#core/components/button-group ?

wadjih-bencheikh18:
> I noticed that when you use the vertical version with the `fill` option to `true`, it fills the horizontal space instead of the vertical space. Can you check if it can be changed to be like https://blueprintjs.com/docs/#core/components/button-group ?

It's working now

stropitek (approved):
Do you see any breaking changes that need to be documented?

Please provide a commit message that summarises the API changes.

wadjih-bencheikh18:
> Do you see any breaking changes that need to be documented?

Only changing the default minimal in toolbar which I added a commit for it

targos:
This commit broke popover items on toolbars: https://github.com/zakodium-oss/react-science/issues/773

PR #740, opened 2024-06-30
chore: update dependencies and fix some ESLint warnings

targos:


src/components/fullscreen/FullscreenContext.tsx
stropitek, line 1: Rename to `.ts` to hint that this file does not / should not contain jsx

src/components/root-layout/RootLayoutContext.tsx
stropitek, line 1: ditto

stropitek (approved):

PR #749, opened 2024-07-23
fix: replace 'button' with 'div' to prevent nested button

hamed-musallam:


src/components/accordion/Accordion.tsx
```diff
@@ -83,9 +83,8 @@ Accordion.Item = function AccordionItem(props: AccordionItemProps) {
         flexDirection: 'column',
       }}
     >
-      <button
+      <div
         onClick={onClickHandle}
```
targos, line 87:
```suggestion
        onClick={onClickHandle}
        role="button"
```

targos (approved):

PR #753, opened 2024-07-30
feat!: replace `getRowTrProps` with render prop `renderRowTr`

targos:
The previous API was not flexible enough. Sometimes we need to call hooks in the context of the rows.
Now this is possible with the pattern:

```
renderRowTr={({ row, children }) => {
  return <CustomTr row={row}>{children}</CustomTr>
}}
```


tpoisseau (approved):
SGTM

Issue #725, opened 2024-05-21
Toolbar dropdown with click interaction should be modal


stropitek:
Meaning it should disable interaction with the rest of the page until it is closed.

If you look at libraries like radiu-ui or spectrum, dropdown menus are modal.

For example in pixelium, I would expect clicking outside of the dropdown to only close the list, not to open the file selector:

https://github.com/zakodium-oss/react-science/assets/4118690/7ab36000-fe3c-4847-8f3c-632eb846f19e

PR #756, opened 2024-08-13
fix: toolbar popover item should have backdrop to avoid interactions outside of it when it is open

stropitek:
Closes: https://github.com/zakodium-oss/react-science/issues/725


stropitek:
@hamed-musallam Assigning you as a reviewer because you might consider this breaking in NMRium.

targos (approved):

hamed-musallam (approved):
LGTM !!! Thanks

PR #780, opened 2024-10-23
Switch from radix to blueprint

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-science/issues/469

For the RadioGroup component, I created the same we have with BluePrintJs
Either we update ours and make it use BluePrintJs instead of radix or just remove the component because it can be done using BluePrintJs

Technically I prefer we update ours because recreating it with BluePrintJs is not simple.

### DIFF
--- a/src/components/info-panel/InfoPanel.tsx
+++ b/src/components/info-panel/InfoPanel.tsx
@@ -1,7 +1,6 @@
 /** @jsxImportSource @emotion/react */
-import { Icon, InputGroup } from '@blueprintjs/core';
+import { Collapse, Colors, InputGroup } from '@blueprintjs/core';
 import { css } from '@emotion/react';
-import * as Collapsible from '@radix-ui/react-collapsible';
 import {
   type CSSProperties,
   memo,
@@ -10,6 +9,7 @@ import {
   useState,
 } from 'react';
 
+import { Button } from '../button/Button.js';
 import { SelectedTotal } from '../selected-total/index.js';
 import { createTableColumnHelper, Table } from '../table/index.js';
 import * as ValueRenderers from '../value-renderers/index.js';
@@ -55,28 +55,27 @@ const style = {
     display: 'flex',
     flexDirection: 'column',
   }),
-  chevron: css({
-    transition: 'all 0.3s ease-in-out',
-  }),
-  button: css({
-    zIndex: 10,
-    position: 'sticky',
-    height: 30,
-    top: 0,
-    "&[data-state='open'] > span": {
-      rotate: '90deg',
-    },
-    cursor: 'pointer',
-    borderBottom: '1px solid #f5f5f5',
-    backgroundColor: 'white',
-    display: 'flex',
-    alignItems: 'center',
-    padding: '5px 2px',
-    width: '100%',
-    ':hover': {
-      backgroundColor: '#f5f5f5',
-    },
-  }),
+  button: (open: boolean) =>
+    css({
+      backgroundColor: `${Colors.LIGHT_GRAY5} !important`,
+      zIndex: 10,
+      position: 'sticky',
+      height: 30,
+      top: 0,
+      '& .bp5-icon': {
+        rotate: open ? '90deg' : '',
+        transition: 'all 0.3s ease-in-out',
+      },
+      cursor: 'pointer',
+      borderBottom: '1px solid #f5f5f5',
+      display: 'flex',
+      alignItems: 'center',
+      padding: '5px 2px',
+      width: '100%',
+      ':hover': {
+        backgroundColor: `${Colors.LIGHT_GRAY4} !important`,
+      },
+    }),
 };
 
 interface InfoPanelDatum {
@@ -209,6 +208,7 @@ interface InfoPanelContentProps {
 }
 
 const InfoPanelContent = memo((props: InfoPanelContentProps) => {
+  const [isOpen, setIsOpen] = useState<string[]>([]);
   const { filteredData } = props;
   return (
     <div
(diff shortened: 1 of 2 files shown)


stories/components/radio.stories.tsx
```diff
@@ -56,6 +59,98 @@ ControlBlueprint.args = {
...
+        '& > label': {
+          color: disabled ? '' : enabledColor,
+        },
+      },
+      fontSize: '1.125em',
+      lineHeight: '30px',
+    }),
+};
+
+export function ButtonRadionBlueprint() {
```
stropitek, line 121: typo

src/components/info-panel/InfoPanel.tsx
```diff
@@ -55,28 +55,27 @@ const style = {
...
-    alignItems: 'center',
-    padding: '5px 2px',
-    width: '100%',
-    ':hover': {
-      backgroundColor: '#f5f5f5',
-    },
-  }),
+  button: (open: boolean) =>
+    css({
+      backgroundColor: `${Colors.LIGHT_GRAY5} !important`,
```
stropitek, line 60: Check the background color, it seems to have changed wrt main.

stropitek (commented):
Make sure to remove any non-used packages.

Do not change anything in the design of the component. I see a background color change on InfoPanel. Double check for other differences.

wadjih-bencheikh18:
@stropitek this PR is ready

stories/components/radio.stories.tsx
```diff
@@ -29,15 +26,12 @@ const ExampleGroup = styled.div`
   gap: 20px;
 `;
 export function ControlBlueprint(
-  props: Omit<
-    BlueprintjsRadioGroupProps,
-    'onChange' | 'selectedValue' | 'children'
-  >,
+  props: Omit<RadioGroupProps, 'onChange' | 'selectedValue' | 'children'>,
 ) {
   const [option, setOption] = useState(options[2]);
```
stropitek, line 31: We can make this example simpler by using the `value` in the state, instead of the whole option.

src/components/forms/radio-button-group/RadioGroup.tsx
```diff
@@ -38,38 +37,39 @@ const rootStyles = {
       },
     }),
 };
-export function RadioGroup(props: RadioGroupProps) {
+export function RadioButtonGroup(props: RadioButtonGroupProps) {
```
stropitek, line 40: My suggestion for this component was that you make the API identical to blueprint's `RadioGroup` component. You would spread the props as is from `RadioGroupButton` to blueprint's `RadioGroup`

stropitek:
The hovered style of radio buttons is missing (vs. on the main branch)

stories/components/radio.stories.tsx
wadjih-bencheikh18: improved

src/components/forms/radio-button-group/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,49 @@
+/** @jsxImportSource @emotion/react */
+import { RadioGroup, type RadioGroupProps } from '@blueprintjs/core';
+import { css } from '@emotion/react';
+
+import { type InputVariant } from '../styles.js';
+
+import { RadioButtonItem } from './ButtonRadioItem.js';
+
+export interface RadioButtonGroupProps extends RadioGroupProps {
+  variant?: InputVariant;
```
wadjih-bencheikh18, line 10: Do I keep this `variant` prop? I think it's better to use `large` prop similar to blueprintjs components

stropitek: I agree, ok to change for a `large` boolean prop.

stropitek (commented):
Please make the height and font size of the radio button component similar to other form components.

- Regular: 30px height and 14px font size (using `rem` unit)
- Large: 40px height and 16px font size (using `rem` unit)

Please provide a commit message with the list of breaking change.

stropitek (commented):
There is a regression in the sticky accordion headers of the info panel.

https://github.com/user-attachments/assets/ff8c51c3-a797-4368-802c-31c14065694b

CleanShot 2024-11-04 at 09 17 35@2x

The height should be 30px / 40px including the borders. It should be the same size next to a text input.

Do your commit messages include all the breaking changes or will you provide a commit message with all the breaking changes in the PR thread?

wadjih-bencheikh18:
> Do your commit messages include all the breaking changes or will you provide a commit message with all the breaking changes in the PR thread?

They Include all breaking changes

wadjih-bencheikh18:
> The height should be 30px / 40px including the borders. It should be the same size next to a text input.

Should `border-radius` also be the same?

stropitek:
> Should `border-radius` also be the same?

No does not have to be

stropitek:
The content does not look centered now (it's too low), and there is a double border on the selected item.

CleanShot 2024-11-05 at 10 52 24@2x

stropitek:
ping @wadjih-bencheikh18

Will you be able to finish this?

stropitek:
@wadjih-bencheikh18 I am taking this over.

stropitek:
Commit message

```
refactor: switch from radix to blueprint

BREAKING CHANGE: The `RadioGroup` component has been renamed to `RadioButtonGroup`. Its interface is now the same as blueprintjs's `RadioGroup` component, which changes how multiple of its props are named and typed.
```

targos:
Shouldn't this uninstall the radix dependency?

stropitek:
Yes they are no longer used, I removed them.

PR #783, opened 2024-10-28
chore: improve activity bar and panel layout

stropitek:
- **feat: improve activity bar layout**
- **docs: rename and move stories**


stropitek:
https://782-improve-design-of-activi.react-science.pages.dev/stories/?path=/story/components-activity-bar-and-panel--activity-toolbar-layout

targos (approved):

PR #796, opened 2024-12-02
feat: virtualize table rows

stropitek:


stropitek:
There is a change in how borders are displayed when `striped` and `bordered` is combined.

I prefer the new one:

Before:
CleanShot 2024-12-02 at 16 55 57

After:
CleanShot 2024-12-02 at 16 55 25

targos:
I don't see the difference, even from afar where I see both at the same time.

src/components/table/table_root.tsx
```diff
@@ -1,38 +1,101 @@
...
+  /**
+   * Set to true to enable sticky header.
+   * When using sticky header with virtualized rows, the `<table>` element will be styled with `display: contents`.
+   * Any style applied through `tableProps.style` will no longer have any effect.
+   */
   stickyHeader?: boolean;
+  /**
+   * Use virtualized rows to optimize rendering.
+   * When virtualizing rows, and in order to prevent layout shifts,
+   * it is recommended that all cells have a fixed width.
```
targos, line 63:
```suggestion
   * it is recommended that all columns have a fixed width.
```

targos (changes requested):
stickyHeader doesn't work after some scrolling with the Virtualized and Custom Tr Render stories.

stropitek:
> I don't see the difference, even from afar where I see both at the same time.

The second one has "borders" (technically it's a `box-shadow`) in between rows.

targos (approved):

PR #798, opened 2024-12-03
remove react popper

Sebastien-Ahkrin:


package-lock.json
```diff
@@ -12,7 +12,7 @@
         "@blueprintjs/select": "^5.3.5",
         "@emotion/react": "^11.13.5",
         "@emotion/styled": "^11.13.5",
-        "@popperjs/core": "^2.11.8",
+        "@floating-ui/react": "^0.26.28",
```
targos, line 15: You can move it to devDependencies, since it's not used by public `components`.

targos (approved):

PR #804, opened 2024-12-06
feat(Table): change sorting icons to use blueprint

Sebastien-Ahkrin:


Sebastien-Ahkrin:
According to the documentation: 

You can determine a custom sorting function on a per-column basis using the sortingFn column option.

By default, there are 6 built-in sorting functions to choose from:
`alphanumeric` - Sorts by mixed alphanumeric values without case-sensitivity. Slower, but more accurate if your strings contain numbers that need to be naturally sorted.
`alphanumericCaseSensitive` - Sorts by mixed alphanumeric values with case-sensitivity. Slower, but more accurate if your strings contain numbers that need to be naturally sorted.
`text` - Sorts by text/string values without case-sensitivity. Faster, but less accurate if your strings contain numbers that need to be naturally sorted.
`textCaseSensitive` - Sorts by text/string values with case-sensitivity. Faster, but less accurate if your strings contain numbers that need to be naturally sorted.
`datetime` - Sorts by time, use this if your values are Date objects.
`basic` - Sorts using a basic/standard a > b ? 1 : a < b ? -1 : 0 comparison. This is the fastest sorting function, but may not be the most accurate.
You can also define your own custom sorting functions either as the sortingFn column option, or as a global sorting function using the sortingFns table option.

I choose to set this for each icons : 
- auto / basic / datetime => sort-asc / desc
- text / textCaseSensitive => sort-alphabetical / sort-alphabetical-desc
- alphanumeric / alphanumericCaseSensitive => sort-numerical / sort-numerical-desc

Otherwise its the base Icon.
Then, for me alphanumeric should be use for numbers. If you have something about that, let me know !

src/components/table/table_header_cell.tsx
```diff
@@ -32,26 +35,47 @@ export function TableHeaderCell<TData extends RowData>(
...
+  >(type)
+    .with('auto', 'basic', 'datetime', () => ({
+      asc: 'sort-asc',
+      desc: 'sort-desc',
+    }))
+    .with('text', 'textCaseSensitive', () => ({
+      asc: 'sort-alphabetical',
+      desc: 'sort-alphabetical-desc',
+    }))
+    .with('alphanumeric', 'alphanumericCaseSensitive', () => ({
```
stropitek, line 76: `alpha` means letter. So it's the alphabetic order that is used here and it should not have the numerical icon.

stories/components/table.stories.tsx
```diff
@@ -45,12 +45,17 @@ const columnHelper = createTableColumnHelper<(typeof table)[number]>();
 const columns = [
   columnHelper.accessor('ocl.idCode', {
     header: 'Molecule',
+    sortingFn: () => {
+      return Math.random() - 0.5;
+    },
+    enableSorting: true,
```
stropitek, line 51: Would not have such a meaningless example in the stories.

stropitek: The `isExpensive` boolean is something you can sort though.

src/components/table/table_header_cell.tsx
```diff
@@ -32,26 +35,47 @@ export function TableHeaderCell<TData extends RowData>(
...
+}
+
+function getSortingIcon<TData extends RowData>(
+  type: SortingFnOption<TData> | undefined,
+): { asc: IconName; desc: IconName } {
+  return match<
+    SortingFnOption<TData> | undefined,
+    { asc: IconName; desc: IconName }
+  >(type)
+    .with('auto', 'basic', 'datetime', () => ({
```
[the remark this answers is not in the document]
Sebastien-Ahkrin: Ok, but we have to keep in mind, if we do not set the sortingFn, basic is the default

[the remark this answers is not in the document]
stropitek: If we don't set it, the value in this part of the code is `undefined` right?

Is so it does not matter that basic strategy will be used internally.

stropitek (approved):

Issue #809, opened 2024-12-06
Remove `lodash`


targos:
Only used in the color picker component. We can probably refactor it to avoid the lib.

PR #812, opened 2024-12-09
chore(ColorPicker): remove lodash

Sebastien-Ahkrin:
Closes: https://github.com/zakodium-oss/react-science/issues/809


Sebastien-Ahkrin:
Is this the best way to remove lodash ?

targos:
Ok for me but with the functions extracted as utilities in https://github.com/zakodium-oss/react-science/tree/809-remove-lodash/src/components/utils

src/components/color-picker/react-color/ColorPicker.tsx
```diff
@@ -1,7 +1,7 @@
-import lodashDebounce from 'lodash/debounce.js';
 import type { CSSProperties } from 'react';
 import { useCallback, useEffect, useRef, useState } from 'react';
 
+import { debounce as internalDebounce } from '../../utils/index.js';
```
targos, line 4: There's no reason to call this "internal". We don't do that with other internal functions.

src/components/utils/debounce.ts
```diff
@@ -0,0 +1,16 @@
+export function debounce<T extends (...args: any[]) => any>(
+  func: T,
+  wait: number,
+): (...args: Parameters<T>) => void {
+  let timeout: NodeJS.Timeout | null;
```
targos, line 5:
```suggestion
  let timeout: ReturnType<typeof setTimeout> | null;
```

It's a bit more verbose but at least we don't leak NodeJS types in frontend code. The typing of setTimeout is annoying when `@types/node` is installed in a project...

targos (approved):

PR #814, opened 2024-12-11
808-move-immer-to-devdependencies

Sebastien-Ahkrin:


src/components/accordion/accordion_context.provider.tsx
```diff
@@ -1,53 +1,125 @@
...
+
+      return [...itemsWithoutChangedItem, newItem];
+    })
+    .with({ type: 'REMOVE' }, ({ payload: title }) => {
+      return previous.filter((element) => element.title !== title);
+    })
+    .with({ type: 'CREATE' }, ({ payload }) => {
+      return [...previous, payload];
+    })
+    .otherwise(() => previous);
```
targos, line 84: We should almost never use `otherwise`, but `exhaustive` instead.

src/components/accordion/accordion_context.provider.tsx
```diff
@@ -1,53 +1,125 @@
...
+): AccordionItemState[] {
+  return match(action)
+    .with({ type: 'CHANGE' }, ({ payload }) => {
+      const { title, isOpen } = payload;
+      const item = getItem(title, previous);
+
+      if (item) {
+        const itemsWithoutChangedItem = previous.filter(
+          (element) => element.title !== title,
+        );
```
stropitek, line 37: I would not put the business logic of identifying an element by its title both in `getItem` and here.

This can be refactored by using a splice or the logic can be moved to the utils.

Also I want to draw your attention to the fact that you changed the order of the elements in this actions and in other actions as well, which was not the case before. This is potentially a regression. Have you checked that we don't rely anywhere on the order?

src/components/accordion/accordion_context.provider.tsx
```diff
@@ -1,53 +1,125 @@
...
+
+      const newItem: AccordionItemState = {
+        ...item,
+        isOpen: !item.isOpen,
+      };
+
+      return [...itemsWithoutChangedItem, newItem];
+    })
+    .with({ type: 'REMOVE' }, ({ payload: title }) => {
+      return previous.filter((element) => element.title !== title);
```
stropitek, line 79: You are doing this a lot. Maybe have a `removeItem` utils that does this.

src/components/accordion/accordion_context.provider.tsx
Sebastien-Ahkrin: I moved everything to utils functions. I checked everything on the Storybook and he looks to works !
I will re-check if we use something about index or other

vite.config.mts
```diff
@@ -25,6 +25,7 @@ export default defineConfig({
     jsx: 'automatic',
   },
   optimizeDeps: {
+    exclude: ['storybook'],
```
stropitek, line 28: Can you try removing cache (`node_modules/.cache`) and clean install of deps?

I'm not able to reproduce the issue that made you add this.

targos: Just `npm ci`. It will remove `node_modules`, including the `.cache` subfolder.

targos:
Keep same reviewer if possible (for better context about previous review)

stropitek (approved):

Issue #815, opened 2024-12-11
Scroll to a row in the Table component


stropitek:
Suggested API and usage:

```ts
function MyTable() {
  const [selectedItem] = useState();
  const scrollToRef = useRef<ScrollToRef>();
  useLayoutEffect(() => {
    // Respond to something like the selected item to initial the scroll
    const index = data.indexOf(selectedItem);
    scrollToRef.current(index);
  }, [selectedItem]);

  return <Table columns={columns} data={data} scrollToRef={scrollToRef} />
}
```

It should work with virtualized and un-virtualized tables.

PR #816, opened 2024-12-12
feat(Table): add scrollToRowRef prop to the Table component

stropitek:
Closes: https://github.com/zakodium-oss/react-science/issues/815

### DIFF
--- a/src/components/table/use_table_scroll.ts
+++ b/src/components/table/use_table_scroll.ts
@@ -1,0 +1,53 @@
+import type { RowData, Table } from '@tanstack/react-table';
+import type {
+  ScrollToOptions as VirtualScrollToOptions,
+  Virtualizer,
+} from '@tanstack/react-virtual';
+import { useEffect, useRef } from 'react';
+
+import type { TableProps } from './table_root.js';
+
+export function useTableScroll<TData extends RowData>(
+  tableProps: TableProps<TData>,
+  table: Table<TData>,
+  virtualizer: Virtualizer<HTMLDivElement, Element>,
+) {
+  const tableRef = useRef<HTMLTableElement>(null);
+  const { scrollToRowRef, virtualizeRows } = tableProps;
+  useEffect(() => {
+    if (scrollToRowRef) {
+      if (virtualizeRows) {
+        scrollToRowRef.current = (
+          id: string,
+          options?: VirtualScrollToOptions,
+        ) => {
+          const sortedRows = table.getRowModel().rows; // Get sorted rows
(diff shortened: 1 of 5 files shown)


stropitek:
Known issue: there can be an offset the size of the header depending on the value of `stickyHeader`.

With virtualization, it works as expected with sticky header. Without the sticky header, it can be fixed by passing the `scrollMargin` option but I'm not sure exposing this option to the component is the right solution.

Without virtualization, it work as expected without the sticky header. `scrollIntoView` is not taking the sticky header into account (I don't have an answer on why or if it can be fixed).

stropitek:
Fixing the issue is not easy. I'd suggest not fixing it right away. If we run into this issue, we can always use the `center` alignment strategy to mitigate the issue.

stories/components/table.stories.tsx
```diff
@@ -187,3 +194,146 @@ export function StyledTable(props: ControlProps) {
...
+    const { rowIndex, buttons } = useScrollButtons();
+    const scrollToRef = useRef<ScrollToRow>();
+    useEffect(() => {
+      if (scrollToRef.current) {
+        scrollToRef.current(String(rowIndex), {
+          behavior: props.scrollBehavior,
+          block: props.scrollBlock,
+        });
+      }
+    }, [rowIndex, props.scrollBehavior, props.scrollBlock]);
```
targos, line 286: Can you make a more "React correct" example? You're not supposed to use an effect to do something in reaction to a user clicking a button. Just do the thing in the click handler.

stropitek: Ok, I'll change. I think I've done it that way intuitively because there are situations where it is required that scrolling happens after a state update is rendered on the screen, for example following an action that changed the data table.

src/components/table/use_table_scroll.ts
```diff
@@ -0,0 +1,53 @@
...
+  tableProps: TableProps<TData>,
+  table: Table<TData>,
+  virtualizer: Virtualizer<HTMLDivElement, Element>,
+) {
+  const tableRef = useRef<HTMLTableElement>(null);
+  const { scrollToRowRef, virtualizeRows } = tableProps;
+  useEffect(() => {
+    if (scrollToRowRef) {
+      if (virtualizeRows) {
+        scrollToRowRef.current = (
```
targos, line 20: Are you sure this is right?
So far I've always seen and used [`useImperativeHandle`](https://react.dev/reference/react/useImperativeHandle) to manipulate refs.
The official docs even mention scrolling to a node as a use case.

stropitek: Yes you are right! This is how refs should be exposed to their parent.

PR #817, opened 2024-12-12
refactor(Button): change css to use styled

Sebastien-Ahkrin:


Sebastien-Ahkrin:
DO NOT MERGE BEFORE https://github.com/zakodium-oss/react-science/pull/814

src/components/button/Button.tsx
```diff
@@ -35,9 +69,11 @@ export function Button(props: ButtonProps) {
     disabled = !tooltipProps.content,
     ...otherToolTipProps
   } = tooltipProps;
+
   const InnerButton = buttonProps.disabled
-    ? BlueprintAnchorButton
-    : BlueprintButton;
+    ? styled(BlueprintAnchorButton)(css(children))
+    : styled(BlueprintButton)(css(children));
```
targos, line 75: Never called `styled()` inside a component. This creates a new component which makes React rerender the whole subtree every time.

src/components/button/Button.tsx
```diff
@@ -44,22 +44,32 @@ const Tag = styled(BlueprintTag)`
...
-  `;
-};
+  return base;
+}
+
+const TooltipAnchorButton = styled(BlueprintAnchorButton)<{
+  children: ReactNode;
+}>((template) => {
+  return getStyledCss(template.children);
+});
```
stropitek: Can't you just do:
```ts
const TooltipAnchorButton = styled(BlueprintAnchorButton)(getButtonStyles)
```
?

Where `getButtonStyles` is `getStyledCss` but adapted to work with props of the Button component.

src/components/button/Button.tsx
```diff
@@ -44,22 +44,32 @@ const Tag = styled(BlueprintTag)`
...
+
+const TooltipButton = styled(BlueprintButton)<{ children: ReactNode }>((
+  template,
+) => {
+  return getStyledCss(template.children);
+});
```
stropitek, line 72: ditto

src/components/button/Button.tsx
```diff
@@ -44,22 +44,32 @@ const Tag = styled(BlueprintTag)`
 // Setting the line-height makes sure regular sized buttons have a
 // height consistent with the input fields and across variants.
 // Using the same line height as in the blueprint docs.
-const css = (children: ReactNode) => {
+function getStyledCss(children: ReactNode) {
+  const base: Record<string, any> = {
```
stropitek, line 48: I think you can do better with the types.

src/components/button/Button.tsx
Sebastien-Ahkrin: I have changed this on my new commit ! this should looks good i think !

stropitek (approved):

PR #819, opened 2024-12-12
refactor(MeasurementsPanel): change css to use styled

Sebastien-Ahkrin:


src/app/panels/measurements/MeasurementsPanel.tsx
```diff
@@ -1,7 +1,7 @@
 /** @jsxImportSource @emotion/react */
 import type { PanelProps } from '@blueprintjs/core';
-import { Tab, Tabs } from '@blueprintjs/core';
-import { css } from '@emotion/react';
+import { Tab, Tabs as BlueprintTabs } from '@blueprintjs/core';
```
stropitek, line 3: I prefer that you use `as` only if 2 imports have the same name. Here you can avoid it by giving a custom name to the styled component.

targos (approved):

stropitek (approved):

Issue #806, opened 2024-12-06
Use styled components instead of `css`


targos:
This means converting uses from `@emotion/react` and the `css` prop to `@emotion/styled`.

@stropitek do you agree?

stropitek:
Yes I agree. I started migrating some of them already while working on features.

When all instances of a file have been refactored, the `/** @jsxImportSource @emotion/react */` comment can be removed.

Sebastien-Ahkrin:
https://github.com/zakodium-oss/react-science/pull/834#discussion_r1888349554

Sebastien-Ahkrin:
This issue is now fixed by every Refs linked here!

PR #824, opened 2024-12-16
refactor(PanelHeader): change css to use styled

Sebastien-Ahkrin:

### DIFF
--- a/src/components/header/PanelHeader.tsx
+++ b/src/components/header/PanelHeader.tsx
@@ -12,7 +12,7 @@ interface PanelHeaderProps {
   children?: ReactNode;
 }
 
-const PanelContainer = styled.div`
+const HeaderContainer = styled.div`
   display: flex;
   flex-direction: row;
   align-items: center;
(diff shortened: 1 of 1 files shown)


src/components/header/PanelHeader.tsx
```diff
@@ -1,49 +1,45 @@
...
   children?: ReactNode;
 }
 
-export function PanelHeader({
-  total,
-  current,
-  children,
-  onClickSettings,
-}: PanelHeaderProps) {
+const PanelContainer = styled.div`
```
targos, line 15: I think this name is confusing because we are not in a Panel component. `PanelHeaderContainer` or `HeaderContainer` or just `Container` would be better, no?
Same below.

targos (approved):

src/components/header/PanelHeader.tsx
Sebastien-Ahkrin: I renamed vars to HeaderContainer & HeaderLeftContainer is this ok @targos ?

targos: yes

PR #833, opened 2024-12-17
docs(Button): change css to use styled

Sebastien-Ahkrin:


stories/components/button.stories.tsx
```diff
@@ -1,6 +1,5 @@
 /** @jsxImportSource @emotion/react */
```
stropitek, line 1: This can be removed once there are no calls to `css` anymore in a file

stropitek (approved):

PR #834, opened 2024-12-17
refactor(Title): change css to use styled

Sebastien-Ahkrin:


src/components/value-renderers/Title.tsx
```diff
@@ -1,21 +1,16 @@
 /** @jsxImportSource @emotion/react */
```
stropitek, line 1: This can be removed once you only use styled components in a file.

stropitek (approved):

src/components/value-renderers/Title.tsx
Sebastien-Ahkrin: This job will be done at the end oh the refactor

PR #838, opened 2024-12-17
docs(PanelHeader): change css to use styled

Sebastien-Ahkrin:


stories/components/panel-header.stories.tsx
```diff
@@ -191,6 +191,10 @@ function PanelContainer({
   );
 }
 
+const PanelStack2Item = styled(PanelStack2)`
```
stropitek, line 194: Why item?

Item is usually a word we use for lists that have several *items*.

stropitek (approved):

PR #843, opened 2024-12-18
docs(FifoLoggerDialog): create story with logger

Sebastien-Ahkrin:
I didn't add any stories with ``FifoLoggerToolbarItem`` because this component is rendered for the "App" application. And it uses the FifoLoggerDialog component.


Sebastien-Ahkrin:
Can be found here: https://9d447fd8.react-science.pages.dev/stories/?path=/story/components-logger--control

stories/components/logger.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+    },
+    onClose: {
+      table: {
+        disable: true,
+      },
+    },
+  },
+  args: {
+    isOpen: true,
+    logs: [
```
stropitek, line 32: Can you add a more representative set of logs, in particular with different levels

stories/components/logger.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+        level: 10,
+        error: new Error('Cannot fetch data'),
+        levelLabel: 'error',
+        message: 'Something went wrong!',
+        uuids: [],
+        meta: {},
+        time: Date.now(),
+      },
+    ],
+    unseen: 1,
```
stropitek, line 44: Make it such that there are seen and unseen logs in the example.

PR #845, opened 2024-12-20
fix: add stylelint and fix CSS errors

targos:


stropitek:
💯

package.json
```diff
@@ -33,7 +33,8 @@
     "prettier-write": "prettier --write .",
     "serve": "vite preview",
     "storybook": "storybook dev -p 6006",
-    "test": "npm run playwright && npm run test-only && npm run eslint && npm run prettier && npm run check-types",
+    "stylelint": "stylelint 'src/**/*.{ts,tsx}' 'stories/**/*.{ts,tsx}'",
```
tpoisseau, line 36: miss css from `/styles` and `/src/pages/demo` at least

src/components/accordion/accordion.tsx
```diff
@@ -32,16 +32,13 @@ const AccordionItemHeader = styled.div`
   cursor: pointer;
   display: flex;
   align-items: center;
-  border-color: rgb(213, 213, 213) currentcolor currentcolor;
+  border-color: #d5d5d5 currentcolor currentcolor;
   border-style: solid none none;
-  color: rgb(45, 45, 45);
+  color: #2d2d2d;
```
tpoisseau, line 37: I'm not convince on default rules about colors.

`rgba(rrr, ggg, bbb, 0.5)` should not transform to `rgb(rrr ggg bbb / 50)`. and `rgb()` should not be transformed to hexadecimal syntax.

Bundling could include this kind of aggressive transformation, but a linter, I'm not sure.

stories/components/tabs.stories.tsx
```diff
@@ -89,10 +89,9 @@ const ManyTabsStyled = styled(Tabs, {
   height: 100%;
 
   div[role='tablist'] {
-    overflow-x: ${(props) =>
-      props.orientation === 'vertical' ? 'hidden' : 'auto'};
-    overflow-y: ${(props) =>
-      props.orientation === 'vertical' ? 'auto' : 'hidden'};
+    overflow: ${(props) =>
+        props.orientation === 'vertical' ? 'hidden' : 'auto'}
+      ${(props) => (props.orientation === 'vertical' ? 'auto' : 'hidden')};
```
tpoisseau, line 94:
```suggestion
    overflow: ${(props) => props.orientation === 'vertical' ? 'hidden auto' : 'auto hidden'};
```

tpoisseau: If we already use `ts-pattern` on this project, It could be more readable than a ternary.

src/components/accordion/accordion.tsx
targos: I did the hexadecimal syntax manually because of the complex background below that took 3 lines with prettier. This is more compact (and I put everything in hexadecimal in this file for consistency)

targos: I don't have strong opinion on the format, but I don't think the goal of the default rules is to do an aggressive transformation for optimization.

This is the error if I put `rgba(r,g,b,0.2)`:

  37:10  ✖  Expected modern color-function notation  color-function-notation
  37:24  ✖  Expected "0.2" to be "20%"               alpha-value-notation

tpoisseau: Lets trust stylelint recommended config so. We will get used to it.

targos:
It is fixed by https://github.com/zakodium-oss/react-science/pull/845/commits/685f23177df25e5868eaf36cbbfec96eadc32e8b

targos:
I let you look again at the diff since you missed it

PR #848, opened 2024-12-27
fix(Menus): do not overflow divider

Sebastien-Ahkrin:
Attention! its a docs not a fix

### DIFF
--- a/stories/components/menus.stories.tsx
+++ b/stories/components/menus.stories.tsx
@@ -12,6 +12,11 @@ const StyledMenu = styled(Menu)`
   border: 1px solid black;
 `;
 
+/*
+ This is a styled version of MenuDivider.
+ By default, Blueprint adds a negative x-axis margin of -5px, which causes the Divider to overflow the Menu on the left and right.
+ We need to override this margin to 0 to prevent it from exceeding the Menu boundaries.
+ */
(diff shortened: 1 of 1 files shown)


stropitek (approved):
I think it would be good to give an explanation.

From what I understand, the MenuDivider component is designed to be rendered within another component which has padding, and therefore has a negative horizontal margin that we need to override.

Sebastien-Ahkrin:
By default (And I don't know how) `MenuDivider` from blueprint has a negative margin. My fix is simply to override this to change -5 to 0

CleanShot 2025-01-03 at 10 40 57

Sebastien-Ahkrin:
Do you want me to add a comment to explain that in the code when i override the style ? @stropitek

stropitek:
> Do you want me to add a comment to explain that in the code when i override the style ?

Yes this is what I meant when I said it would be good to have an explanation.

Sebastien-Ahkrin:
I added this : 

/*
 This is a styled version of MenuDivider.
 By default, Blueprint adds a negative x-axis margin of -5px, which causes the Divider to overflow the Menu on the left and right.
 We need to override this margin to 0 to prevent it from exceeding the Menu boundaries.
 */

If we are ok, i can merge this !

PR #851, opened 2025-01-02
feat(InputPercent): add InputPercent component

Sebastien-Ahkrin:


stropitek:
Closing this, it's not worth adding something we don't actually need since we cannot even define how it is supposed to work.

Issue #859, opened 2025-02-19
[Blueprint] <Popover> targetProps value is ignored when renderTarget API is used.


targos:
I'm seeing a lot of warnings related to this in NMRium.

They come from `ToolbarPopoverItem`

https://github.com/zakodium-oss/react-science/blob/6cf1bf98f435c84b95a410c583c24684586ffda4/src/components/toolbar/Toolbar.tsx#L254-L265

@stropitek can you take this?

PR #860, opened 2025-02-19
fix: remove targetProps where renderTarget is used

stropitek:
Blueprintjs prints a warning about the prop having no effect in that case.

Closes: https://github.com/zakodium-oss/react-science/issues/859

### DIFF
--- a/src/components/toolbar/Toolbar.tsx
+++ b/src/components/toolbar/Toolbar.tsx
@@ -50,7 +50,8 @@ interface ToolbarItemInternalProps extends ToolbarItemProps {
   isPopover: boolean;
 }
 
-export interface ToolbarPopoverItemProps extends PopoverProps {
+export interface ToolbarPopoverItemProps
+  extends Omit<PopoverProps, 'targetProps'> {
   itemProps: ToolbarItemProps;
 }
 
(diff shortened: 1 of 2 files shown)


src/components/toolbar/Toolbar.tsx
```diff
@@ -251,13 +252,6 @@ Toolbar.PopoverItem = function ToolbarPopoverItem(
       interactionKind={popoverInteractionKind}
       hasBackdrop={popoverInteractionKind === 'click'}
       hoverCloseDelay={0}
-      targetProps={{
-        style: {
-          fontSize: '1.125em',
-          width: 'fit-content',
-          height: 'fit-content',
-        },
-      }}
```
targos, line 260: Then this was useless? I wonder why it was there in the first place.

targos (approved):

src/components/toolbar/Toolbar.tsx
stropitek: I think I just forgot to remove it when I changed to implementation to use `renderTarget`

[fix: ensure popover toolbar item does not grow](https://github.com/zakodium-oss/react-science/commit/9d8a0b459c813295af29cf90200e35540b495840)

Issue #852, opened 2025-01-07
improve split panel


hamed-musallam:
1. Add an `interactionKind` Property
- Current issue:
When a user interacts with the splitter (e.g., clicks the "..." button), the closed property no longer controls the panel's open/close state.

- Proposed solution:
Introduce an optional property called `interactionKind`, which can take one of two values:
    1- forced: Ensures the panel behavior is fully controlled programmatically.
    2- undefined: Allows default behavior without restrictions.

2. Refactor the `closed` property:
- Current issue:
The closed property is overloaded with two types of values:
Boolean: Directly controls whether the panel is open or closed.
String: Represents a size threshold for automatic closing.

- Proposed solution:
The closed property should only accept a boolean value for controlling the open/close state.
Introduce a new property (e.g.,  `closeThreshold`) to handle automatic closure based on panel size.

3. Add an `onClose` Callback:
- Current issue:
When using the size threshold for automatic closing, there’s no way to track whether the panel is currently closed.

- Proposed solution:
Add an `onClose` callback function that triggers whenever the panel is closed, regardless of whether it was closed manually or automatically.

4. Improve automatic close behavior:
 - The automatic closing feature should always be active, based on the `closeThreshold`. I think we need to differentiate between two scenarios:
        1- User Interaction: When the user changes the panel size using the "..." button.
        2- Container Resizing: If the container size causes the panel to shrink below the threshold, it should automatically close.

- https://github.com/cheminfo/nmrium/pull/3318

stropitek:
> 1. Add an `interactionKind` Property
> 
> * Current issue:
>   When a user interacts with the splitter (e.g., clicks the "..." button), the closed property no longer controls the panel's open/close state.
> * Proposed solution:
>   Introduce an optional property called `interactionKind`, which can take one of two values:
>   1- forced: Ensures the panel behavior is fully controlled programmatically.
>   2- undefined: Allows default behavior without restrictions.

I prefer that we use this typical react pattern:
- If the property is undefined, it is uncontrolled (which means automatically controlled by the component)
- If the property is defined, it is controlled. We must introduce a callback prop (something like `onClosedChange`) which is called when we click the separator or when the closed value changes for any other reason (like the example with the `closeThreshold`).
- The component is not supposed to switch from being uncontrolled to controlled and vice versa (print a warning if that happens).

> 2\. Refactor the `closed` property:

SGTM


> 3\. Add an `onClose` Callback:

[...]

PR #862, opened 2025-02-19
feat: improve split panel

stropitek:
Commit message:
```
feat: rework SplitPane API

Refactor the SplitPane API to be more consistent and idiomatic.
Make the `open` and `size` state of the component controllable.

BREAKING CHANGE: The `onToggle` prop has been renamed to `onOpenChange`
and receives true when the split pane opens (previously received false). 
Contrary to the previous onToggle, onOpenChange is called in any
circumstance when the split pane state changes, for example also when it
automatically closes after it reaches the `closeThreshold` size.
The original behavior of the `closed` and `size` props has changed.
Their behavior were a mix of a control value prop and default value prop
for internally managed state.
There is now a clear separation between control props (`open` and `size`)
and default value props `defaultOpen` and `defaultSize`.
For example for the open state you should either:
- Use `open` to control if the SplitPane is open or not, and `onOpenChange` to react to changes. Using `defaultOpen` will have no effect.
- Use `defaultOpen` to set the initial value (only affects first render), with state changes being handled internally by the component.
If you used `closed` with a number, use the new `closeThreshold` prop instead.


Closes: https://github.com/zakodium-oss/react-science/issues/852
```

Closes: https://github.com/zakodium-oss/react-science/issues/852


src/components/split-pane/SplitPane.tsx
```diff
@@ -111,48 +134,42 @@ export function SplitPane(props: SplitPaneProps) {
...
+      setSize(serialized);
+      onSizeChange?.(serialized);
     },
     onResize,
   });
 
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore Module exists.
+  // @ts-expect-error I do not understand why there is a TS error here.
   const rootSize = useResizeObserver<HTMLDivElement>();
```
stropitek, line 145: @targos could you have a look. I don't understand why there is an error.

src/components/split-pane/SplitPane.tsx
```diff
@@ -111,48 +134,42 @@ export function SplitPane(props: SplitPaneProps) {
...
+      const serialized = serializeSize(value);
+      setSize(serialized);
+      onSizeChange?.(serialized);
     },
     onResize,
   });
 
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore Module exists.
+  // @ts-expect-error I do not understand why there is a TS error here.
```
stropitek, line 144: @targos could you have a look?

I don't understand why there is an error.

targos: I don't understand either, but this is why I used `@ts-ignore` in this case.

src/components/utils/controlled_or_internal_state.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (value === undefined && previousValue !== undefined) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        `SplitPane: "${name}" prop changes from being controlled to being uncontrolled. This may lead to unexpected behavior.`,
+      );
+    }
+  }, [value, previousValue, name]);
+}
+
+export function useControlledOrInternalState<T>(
```
targos, line 31: In another project, we depend on `@radix-ui/react-use-controllable-state`. Have you considered it?

src/components/utils/controlled_or_internal_state.ts
```diff
@@ -0,0 +1,49 @@
+import { useCallback, useEffect, useRef, useState } from 'react';
+
+function usePrevious<T>(value: T) {
+  const ref = useRef<T>(value);
+
+  useEffect(() => {
+    ref.current = value;
+  }, [value]);
+
+  return ref.current;
```
targos, line 10: This is forbidden by React (and I think incompatible with React 19 or the React Compiler).

targos (approved):
Behavior and API LGTM

src/components/utils/controlled_or_internal_state.ts
stropitek: I had a vague memory about it but did not know where to look for.
I will try to use that.

src/components/utils/controlled_or_internal_state.ts
stropitek: I removed that part after using the suggestion radix-ui hook.

Issue #868, opened 2025-02-27
Reorder table rows


stropitek:

stropitek:
Major question mark is what library to use or if we prefer not to use a library.

That question should be considered in the context that react-science / NMRium has multiple potential applications for DnD interface. The ones I can think of are:
- Reorder table rows or columns
- Reordering panels
- Molecule assignment

Currently NMRium has user interface to drag things around like molecule, peak peaking annotation, title block, but no library is used for that (there is no drop target and the logic is very simple).

I've explored a bit the `@dnd-kit` packages:

Pros:
- Featured in examples in react-table docs
- We've used it in a project and haven't had issues with it
- Helpers for sorting out of the box. I will be relatively straightforward to integrate in the Table component.

Cons:
- The library is currently in between a major upgrade to framework agnostic approach. See https://github.com/clauderic/dnd-kit/issues/1194. It is currently in beta, is actively worked on, but it will probably be a while until there is a stable release. Docs of the beta version: https://next.dndkit.com/overview
- A lot people have been complaining about too many re-renders, which is one of the things the new major version is supposed to address. I'm not too worried about this since we have implemented virtualization in tables.

Other candidates I haven't looked into very much but seem valid:
- https://github.com/atlassian/pragmatic-drag-and-drop
- https://github.com/hello-pangea/dnd

@zakodium-oss/collaborators let me know if you have any opinions.

PR #872, opened 2025-03-04
feat(Table): allow to pass custom td attributes and reorder rows

stropitek:
- Add `getTdProps` prop which allows to customize the attributes passed to the `<td>` elements, including style. This is useful for example to center content within a cell as this can not easily be done in tables based on the cell's children only.
- Add `onRowOrderChanged` callback which is called when a manual reordering operation is finished, with the data items in their new order.
- Add `renderRowPreview` to customize the content of the dragged row's preview.
- Add `scrollableElementRef` prop which allows to pass an external reference to the element that should be auto-scrolled when reordering rows.
- Export new `TableDragRowHandler` for a good default drag handler that can be used in cell renderers.

Other new hooks and components which allow to customize the row reordering UI:
- `useTableDraggableRowContext` and `TableDropIndicator` for implementing your own drag handler instead of `TableDragRowHandler`.
- `TableDraggableRowTr` component which can be used in custom row rendering with support of row reordering.

When the reordering callback is used, there is a warning if:
- some columns are sortable as sorting is incompatible with reordering
- `getRowId` is not specified, since rows must have stable identifiers during reordering.

There was some research on different library candidates: https://github.com/zakodium-oss/react-science/issues/868#issuecomment-2688261956
The atlaskit library was retained because of support from a big company + it being lightweight.

BREAKING CHANGE: td cells no longer have `position: relative;` by default but this can be overridden using `getTdProps`.

Closes: https://github.com/zakodium-oss/react-science/issues/868

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,9 @@
       "version": "14.0.0",
       "license": "MIT",
       "dependencies": {
+        "@atlaskit/pragmatic-drag-and-drop": "^1.5.0",
+        "@atlaskit/pragmatic-drag-and-drop-auto-scroll": "^2.1.0",
+        "@atlaskit/pragmatic-drag-and-drop-hitbox": "^1.0.3",
         "@emotion/styled": "^11.14.0",
         "@radix-ui/react-use-controllable-state": "^1.1.0",
         "@tanstack/react-table": "^8.21.2",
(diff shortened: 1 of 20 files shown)


stropitek:
There is a warning about the `large` prop of `Tag` being deprecated but I am not using it...

tpoisseau (approved):
SGTM. Is it possible (in an other PR) to have the complete row preview during the drag action ? It would be the best default behavior.

stropitek:
FYI I checked how much space the atlaskit libraries were taking in the demo app bundle and it is about 90kb uncompressed (more than I expected).

PR #882, opened 2025-03-06
SplitPane with initial size of 0

stropitek:


stropitek:
Not able to reproduce the issue I had on peak-by-peak with a minimalistic example.

PR #887, opened 2025-03-10
feat: rework Accordion API

stropitek:
Squash commit message:

```txt
feat: rework Accordion API and add new features to it

New controlled / uncontrolled API with `open` and `defaultOpen` props.
New `closeAllExcept` and `toggle` methods on `useAccordionControls` hook.

Remove padding where accordion item's toolbar is placed. The toolbar should decide of it's own padding or margin instead.

BREAKING-CHANGE: The `useAccordionContext` hook has been removed from the API. It should only be used internally.
The `defaultOpened` prop of `Accordion.Item` has been renamed to `defaultOpen`.
The `title` prop no longer acts as an identifier, and an additional `id` props has been added to that effect.
The `useToggleAccordion` hook has been renamed to `useAccordionControls`.
Using the same identifier twice now throws a global JS error (not a render error).
The `toolbar` prop has been changed to `renderToolbar` to allow customization of render base on open state.

Closes: https://github.com/zakodium-oss/react-science/issues/876
```


stories/components/accordion.stories.tsx
```diff
@@ -337,16 +340,91 @@ export const UnmountChildren = {
...
+        title="Second"
+        open={openItems.includes('Second')}
+        onOpenChange={(isOpen) =>
+          isOpen ? addItem('Second') : removeItem('Second')
+        }
+      >
+        Second content
+      </Accordion.Item>
+    </Accordion>
+  );
```
tpoisseau, line 413: I'm not convinced by the controlled state api. It is verbose and repetitive.

I think the state should be managed at `Accordion` level, with `opens` / `defaultOpens` / `onChange` props.

stropitek: The main use case is to use the "uncontrolled" version and use the `useAccordionToggle` hook if we need to control the opened / closed items from outside.

The `defaultOpens` prop on `Accordion` would make it more difficult to use it for this use case (need to repeat the `id` in 2 different components).

I agree that the story is verbose, but in a real case scenario of controlled state we would have a simple `dispatch` on change and a wrapper component around `Accordion.Item` to make things much less verbose.

src/components/accordion/accordion.tsx
```diff
@@ -8,18 +9,37 @@ import { useCallback } from 'react';
...
 }
 
-export interface AccordionItemProps {
-  title: string;
+export interface AccordionItemProps<T extends string = string> {
+  /**
+   * The tile of the accordion item.
+   * The title also acts as an identifier and should be unique within an accordion provider.
+   */
+  title: T;
```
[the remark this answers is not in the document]
stropitek: ok I will do that (I though about it but in the end tried not to make too many breaking changes)

src/components/accordion/accordion_context_provider.tsx
```diff
@@ -0,0 +1,80 @@
...
+      unregister: (title: string) => {
+        assert(registerRef.current);
+        registerRef.current.delete(title);
+      },
+      register: (title: string, setIsOpen: AccordionItemSetIsOpen) => {
+        assert(registerRef.current);
+        const item = registerRef.current.get(title);
+        if (item) {
+          // eslint-disable-next-line no-console
+          console.error(
```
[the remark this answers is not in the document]
stropitek: Ok I will do an assertion then.

The error can only happen in a `useEffect`, so it will result in an uncaught exception and not catched by error boundaries.

stropitek:
@tpoisseau I made some additional API changes unrelated to your comments compared to when you last reviewed.

stropitek:
See https://github.com/zakodium-oss/react-science/pull/887#issue-2907794350 for an overview of the changes.

stories/components/accordion.stories.tsx
tpoisseau: Yes, please provide a story which control the state with `useAccordionToggle`

tpoisseau (approved):
SGTM, a better story for control state would be great.

stories/components/accordion.stories.tsx
stropitek: New story using a reducer + context to manage the external state provided.

PR #890, opened 2025-03-13
feat(accordion): highlight interactivity by changing the header's color

stropitek:
And use gray colors from the blueprint library


targos:
Have you tried different approaches for the style? This inversion of the gradient feels odd when I test it.

stropitek:
> Have you tried different approaches for the style? This inversion of the gradient feels odd when I test it.

I tried without the inversion and it was not bad but more subtle. I push another commit so that you can try. I'm fine with either.

targos (approved):
I prefer like that.

PR #892, opened 2025-03-18
feat: render entire row as drag preview

stropitek:
- **feat(Table): add noHeader prop**
- **feat(Table): render dragged row as the preview by default**


tpoisseau (approved):
Is it possible to vertical center the preview, relative to the mouse cursor ?

src/components/table/table_root.tsx
```diff
@@ -28,30 +31,33 @@ import { useTableColumns } from './use_table_columns.js';
...
-          props.bordered
-            ? 'inset 1px 0 0 0 #11141826 !important'
-            : 'none !important'
-        };
+        box-shadow: ${match(props)
+          .with(
+            { bordered: true },
+            () => 'inset 1px 0 0 0 #11141826 !important',
+          )
+          .otherwise(() => 'none !important')};
```
Sebastien-Ahkrin, line 53: Its a bit strange for me to use ts-pattern only for a ternary

src/components/table/table_root.tsx
```diff
@@ -28,30 +31,33 @@ import { useTableColumns } from './use_table_columns.js';
 import { useTableScroll } from './use_table_scroll.js';
 
 const CustomHTMLTable = styled(HTMLTable, {
-  shouldForwardProp: (prop) => prop !== 'striped' && prop !== 'stickyHeader',
-})<{ stickyHeader: boolean }>`
+  shouldForwardProp: (prop) =>
+    prop !== 'striped' && prop !== 'stickyHeader' && prop !== 'noHeader',
```
Sebastien-Ahkrin, line 35: This start to be hard to understand maybe we can refactor this to a function or something else ?

Sebastien-Ahkrin (approved):

stropitek:
> LGTM, I just find it strange that the drag & drop button is so small. Maybe try to make it bigger / take up the entire cell ?

Maybe it's because in this particular case, the row takes a lot of height because of the molecule? 

The drag handler is not necessary meant to be used standalone in a cell. It can be just a part of a cell. My implementation is based on https://atlassian.design/components/pragmatic-drag-and-drop/examples, with a similar sized drag handler. It is true though that mine is slightly smaller.

src/components/table/table_root.tsx
stropitek: Initially thought I needed more than 2 cases.

But not addressing this comment as I think it's a matter of personal preferences.

stropitek:
I now forward props from the drag handler component to the button component to be able to customize it.

Issue #891, opened 2025-03-17
Add flashing effect when scrolling to table row


stropitek:
Similar to the flashing effect when we drag and drop rows.

PR #895, opened 2025-03-18
feat: add flashing effect when scrolling to table row

stropitek:
Closes: https://github.com/zakodium-oss/react-science/issues/891


stropitek:
Sorry if the PR is messy because of the refactoring.

But essentially this is reusing some existing behavior for flashing lines in a table that was implemented for reordering, and applying this when scrolling a row into view as well.
It is optional and opt in to flash the row when scrolling into view.

stropitek:
No breaking changes except a renamed type.

targos (approved):

Issue #689, opened 2024-03-16
Add fullscreen provider to root layout


targos:
If possible, without an additional `div`. Currently we have to wrap rootlayout with fullscreen so the dialogs are not open outside of it. See https://github.com/zakodium-oss/react-science/blob/e8aebd40319dcdf4127eb591a6eb952907fc251d/src/pages/demo/App.tsx#L13-L21

PR #897, opened 2025-03-19
feat: change FullScreenProvider API to not create additional div

stropitek:
BREAKING-CHANGE: FullScreenProvider now takes a render callback with the ref to attach to the element which can be displayed fullscreen. It no longer renders an additional styled div.
The RootLayout is now a fullscreenable element. It is no longer necessary to wrap it in a FullScreenProvider. Added a global CSS rule to make fullscreen elements' backdrop background color white.

Closes: https://github.com/zakodium-oss/react-science/issues/689

Note: It is not possible to add a margin to a fullscreen element so the stories are slightly different now.

I find it to make more sense to put the border on the element which is fullscreenable rather than a child of it.
Therefore there is no space outside the border anymore.

### DIFF
--- a/src/components/fullscreen/fullscreen_context.provider.tsx
+++ b/src/components/fullscreen/fullscreen_context.provider.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable no-alert */
-import type { ReactNode } from 'react';
+import type { ReactNode, RefObject } from 'react';
 import { useEffect, useMemo, useReducer, useRef } from 'react';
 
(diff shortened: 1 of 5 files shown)


stropitek:
Forgot to test the App which has a background color issue.

stropitek:
Ready

targos (approved):

stropitek:
Will make a non-breaking release when https://github.com/zakodium-oss/react-science/pull/895 is merged, and then a new release with this breaking change, to allow for incremental upgrades.

Issue #899, opened 2025-03-20
pragmatic-drag-and-drop imports are not compatible with ESM


targos:
If you try to import `react-science/ui` from an ESM module, it fails:

```
import { Button } from 'react-science/ui';
```

Or in node's REPL: `import('react-science/ui')`

Can also be reproduced in `react-science`:

```
npm run build-ts
node lib/components/index.js
```

Error looks like:

```
Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import '/Users/mzasso/git/zakodium-oss/react-science/node_modules/@atlaskit/pragmatic-drag-and-drop/combine' is not supported resolving ES modules imported from /Users/mzasso/git/zakodium-oss/react-science/lib/components/table/reorder_rows/draggable_row_tr.js
Did you mean to import "@atlaskit/pragmatic-drag-and-drop/dist/cjs/entry-point/combine.js"
```

This is related to https://github.com/atlassian/pragmatic-drag-and-drop/issues/27

PR #901, opened 2025-03-24
fix: make sure build is natively compatible with ESM

stropitek:
Given how @atlaskit/pragmatic-drag-and-drop publishes ESM in a non-native way, we have to trick TS into using the CJS build so that our build is compatible with native ESM.

Closes: https://github.com/zakodium-oss/react-science/issues/899

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -11,7 +11,7 @@
     "forceConsistentCasingInFileNames": true,
     "allowJs": false,
     "isolatedModules": true,
-    "verbatimModuleSyntax": true,
+//    "verbatimModuleSyntax": true,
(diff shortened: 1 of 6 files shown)


tsconfig.json
```diff
@@ -11,7 +11,7 @@
     "forceConsistentCasingInFileNames": true,
     "allowJs": false,
     "isolatedModules": true,
-    "verbatimModuleSyntax": true,
+//    "verbatimModuleSyntax": true,
```
targos, line 14:
```suggestion
    "verbatimModuleSyntax": false,
```

targos (approved):

PR #902, opened 2025-03-27
feat(Accordion): improve renderToolbar API

stropitek:
Ref: https://github.com/zakodium/peak-by-peak/issues/360

BREAKING-CHANGE: The render callback parameter has changed to contain multiple properties instead of just the `isOpen` state. It can now be used to control the state of the item in which it is rendered. Click events in the toolbar no longer propagate to the header in order to prevent it from toggling if it is not wanted.

### DIFF
--- a/src/components/accordion/accordion.tsx
+++ b/src/components/accordion/accordion.tsx
@@ -6,9 +6,10 @@ import type {
   ReactElement,
   ReactNode,
 } from 'react';
-import { useCallback } from 'react';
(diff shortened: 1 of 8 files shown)


stropitek:
> Is it possible to have a story combining panel with accordion and activity bar, like Peak-by-Peak ?

The story I changed already shows it works because if the click was propagated, it would call `toggle` twice and not open.

You you want me to add the story anyway?

tpoisseau (approved):
Yes please

PR #903, opened 2025-03-27
feat: improve fullscreen error handling

targos:
> Use a library to manage fullscreen APIs.
> Also add a `onToggleError` prop to the `FullScreenProvider` and `RootLayout` components.
>
> There is a behavior change with "stacked" fullscreen elements (like in the storybook). Now when you enter fullscreen a second time, it will first exit the previous fullscreen. I think it's fine. We don't (and probably shouldn't) use stacked fullscreens in real apps.

I'm not sure about the `onToggleError` prop. Another option is to make the `toggle()` function async, but then the user has to handle it otherwise it will be an unhandled rejection (which for example will trigger error monitors like Sentry).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
         "@tanstack/react-virtual": "^3.13.2",
         "d3-scale-chromatic": "^3.1.0",
         "react-dropzone": "^14.3.8",
+        "react-full-screen": "^1.1.1",
         "react-icons": "^5.5.0",
         "react-inspector": "^6.0.2",
         "tinycolor2": "^1.6.0",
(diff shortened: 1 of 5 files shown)


targos:
For reference, this is what I get with the story on Firefox for iPad: 
CleanShot 2025-03-27 at 13 25 45

stropitek (commented):
The behavior of nested fullscreen element has changed.

It always exits the parent before entering the child.

It creates a strange transition and changes what happens when you exit a nested fullscreen element (does not toggle back to previously fullscreen parent).


https://github.com/user-attachments/assets/e0ba65e8-11ca-4cc4-8627-7dd1aaebd017

> I'm not sure about the `onToggleError` prop. Another option is to make the `toggle()` function async, but then the user has to handle it otherwise it will be an unhandled rejection (which for example will trigger error monitors like Sentry).

SGTM. Maybe create an unhandled rejection only when `onToggleError` is not specified?

targos:
@stropitek Please read my change description again :)

targos:
> Maybe create an unhandled rejection only when onToggleError is not specified?

In that case, I'd rather remove the callback. I added it so that if we do nothing there is no unhandled rejection.

stropitek:
> Please read my change description again :)

oups I got sidetracked midway while reading the commit message.

stropitek (approved):

PR #906, opened 2025-03-27
fix(Table): remove false positive warning about conflicting props

stropitek:
Ref: https://github.com/zakodium/peak-by-peak/issues/375


targos (approved):
That's why I didn't merge the release PR 😄

stropitek:
> That's why I didn't merge the release PR 😄

Yes, I realized a few seconds after I merged it 😆

PR #908, opened 2025-03-27
docs: using react-resizable-panels for activity panels

stropitek:


stropitek:
POC done in story.

We can think of how to integrate this deeper in the API later.

stropitek:
It's not easy to initialize the sizes properly when a panel is added / removed, and we have a configuration of visible panels that we have never seen before.

The default is to split them evenly but this is not what we want. We'd rather have them take space that is a function of the previous configuration but recalculated with one panel more / less.

PR #909, opened 2025-04-01
fix: correct incompatibilities when blueprint stylesheet is in a css layer

stropitek:
- **fix: correct toolbar button when blueprint css is in a layer**
- **refactor: simplify markup for activity item**
- **fix: make sure the toolbar buttons are properly centered when blueprint styles are in a layer.**

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -7,7 +7,6 @@ import type {
 import {
   AnchorButton,
   Button as BlueprintButton,
-  Classes,
   Icon,
   Tag,
   Tooltip,
@@ -15,7 +14,6 @@ import {
 import type { CSSObject } from '@emotion/styled';
 import styled from '@emotion/styled';
 import type { ReactNode } from 'react';
-import { Fragment } from 'react';
 
 type BlueprintProps = {
   [key in keyof BlueprintButtonProps &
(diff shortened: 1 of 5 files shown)


src/components/button/Button.tsx
```diff
@@ -46,22 +44,17 @@ const ButtonTag = styled(Tag)`
 // Setting the line-height makes sure regular sized buttons have a
 // height consistent with the input fields and across variants.
 // Using the same line height as in the blueprint docs.
-function getStyledCss(
-  buttonProps: BlueprintAnchorButtonProps | BlueprintButtonProps,
-): CSSObject {
-  return {
-    lineHeight: 1.15,
-    position: 'relative',
-    [`.${Classes.ICON}`]: !buttonProps.children && { marginRight: 0 },
```
stropitek, line 55: This line is what was creating an incompatibility when using CSS layers for the blueprintjs stylesheets.

It would take precedence over blueprintjs styles because not in a layer.

src/components/toolbar/Toolbar.tsx
```diff
@@ -185,32 +178,28 @@ function ToolbarItemInternal(props: ToolbarItemInternalProps) {
...
+                size={9}
+                style={{
+                  transform: 'rotate(45deg)',
+                  position: 'absolute',
+                  bottom: 0,
+                  right: 0,
+                  width: 9,
+                  height: 9,
+                }}
+              />
```
Sebastien-Ahkrin, line 197: Do this is possible to create the Icon with styled ?

stropitek: It's easier not to use styled here because the width and height need to override the width and height set by the `bp5-icon` class.

Issue #896, opened 2025-03-18
Add helper to handle styled component forwarded props


stropitek:
Maybe what we need is a helper to generate the `shouldForwardProp` function from an array of prop names (that could be converted internally to a Set?

_Originally posted by @targos in https://github.com/zakodium-oss/react-science/pull/892#discussion_r2001408629_

Sebastien-Ahkrin:
I saw your message too late

stropitek:
If you do a wrapper of `styled` I think you should:
- Not change the contract of the function except for refining the types of the `shouldForwardProp` option.
- Find a better name than prefix with `zak`
- Do not have to pass `typeof Dialog` explicitely (it can be inferred from the first argument of the function).

PR #911, opened 2025-05-12
feat: create shouldForwardPropExcept utils

Sebastien-Ahkrin:
Closes: https://github.com/zakodium-oss/react-science/issues/896

### DIFF
--- a/src/components/dialog/ConfirmDialog.tsx
+++ b/src/components/dialog/ConfirmDialog.tsx
@@ -1,8 +1,9 @@
-import { Classes, Dialog, DialogBody, DialogFooter } from '@blueprintjs/core';
 import type { DialogProps } from '@blueprintjs/core';
+import { Classes, Dialog, DialogBody, DialogFooter } from '@blueprintjs/core';
 import styled from '@emotion/styled';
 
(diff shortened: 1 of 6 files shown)


stropitek (commented):
Do you have a suggestion on how to make this better regarding types?

Sebastien-Ahkrin:
> Do you have a suggestion on how to make this better regarding types?

regarding types ?

Sebastien-Ahkrin:
We can go with this API : 

```ts
interface ContainerType extends HTMLAttributes<HTMLDivElement> {
  color: string;
}

const Container = styled('div', {
  shouldForwardProp: shouldForwardPropExcept<ContainerType>(['color']),
})<ContainerType>`
  background-color: ${(props) => props.color};
`;

export function Default() {
  return <Container color="red">Hello, World!</Container>;
}
```

If you are Ok. Its a bit "hard" since shouldForwardProp is typed to give a string as prop

stropitek (approved):
I'm okay with this version as it is not less type safe than before.

Implementing a type safe wrapper of `styled` is not a requirement to merge this PR.

PR #920, opened 2025-06-10
chore: migrate to storybook 9

Sebastien-Ahkrin:
Closes: https://github.com/zakodium-oss/react-science/issues/917


stropitek (commented):
SGTM

I think there are stories using the `autodocs` tag but it does not have any effect (already the case on main). Using the tag filter those ones seem to have the tag:

CleanShot 2025-06-10 at 11 13 04@2x


I noticed we lost the interactions panel but I don't think we have any interactive stories.

PR #928, opened 2025-07-11
927 splitpane do not render side when size is 0

stropitek:
- **feat(SplitPane): add prop to unmount any invisible split pane side**
- **docs: add an example which uses a size in pixels, not percent**


stropitek:
I changed my mind on the default being to unmount children.

This way we have a non-breaking change and an API which is the same as for the `Accordion` component.

PR #932, opened 2025-08-19
feat: integration of tanstack form with our form element components

Sebastien-Ahkrin:


src/components/ts_form/ts_input.tsx
```diff
@@ -0,0 +1,43 @@
...
+import type { ChangeEvent } from 'react';
+import { useCallback, useMemo } from 'react';
+
+import { useFieldContext } from './context/use_ts_form.js';
+
+type TSInputProps = Omit<InputGroupProps, 'defaultValue' | 'name'>;
+
+const Input = styled(InputGroup)<{ error: boolean }>`
+  .${Classes.INPUT} {
+    border: 1px solid ${(props) => (props.error ? 'red' : 'inherit')};
```
stropitek, line 13: You should use Blueprint's intent properties to style the input and avoid hardcoding colors. If you absolutely need to use a color like this, then you should import the color from Blueprint. But it is probably not needed.

src/components/ts_form/ts_input.tsx
```diff
@@ -0,0 +1,43 @@
...
+    [field],
+  );
+
+  const error = useMemo<string | undefined>(() => {
+    const error = field.state.meta.errors.find((e) => e.path[0] === field.name);
+    return error?.message || undefined;
+  }, [field.name, field.state.meta.errors]);
+
+  return (
+    <FormGroup color="red" helperText={error}>
```
stropitek, line 33: `color="red"` ??

stropitek (commented):
There are warning in the console when loading both stories:

CleanShot 2025-08-20 at 10 46 48@2x

---

I don't like the "TS" prefix. I understand it is for "Tanstack" but I read "TypeScript".

stories/components/ts_form.stories.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+        <form.AppField name="lastName">
+          {(field) => (
+            <FormGroup
+              label="Last name"
+              labelFor="lastName"
+              labelInfo={<span style={{ color: 'red' }}>*</span>}
+              intent="danger"
+            >
+              <field.TSInput id="lastName" />
```
stropitek, line 86: Here we are already in the context of a tanstack form field and the "TS" prefix is not needed.

stories/components/ts_form.stories.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+      <form.Subscribe selector={(state) => state.isDirty}>
+        {(isDirty) => (
+          <Alert>Your form {isDirty ? 'is' : "isn't"} dirty.</Alert>
+        )}
+      </form.Subscribe>
+
+      <form.AppForm>
+        <FormWrapper>
+          <form.TSSubmitButton>Submit</form.TSSubmitButton>
```
stropitek, line 100: Ditto for the TS prefix

stories/components/ts_form.stories.tsx
```diff
@@ -0,0 +1,198 @@
...
+      <FormWrapper>
+        <form.AppField name="firstName">
+          {(field) => (
+            <FormGroup
+              label="First name"
+              labelFor="firstName"
+              labelInfo={<span style={{ color: 'red' }}>*</span>}
+              intent="danger"
+            >
+              <field.TSInput id="firstName" />
```
stropitek, line 73: It's a bit confusing to pass this id property because it seems important in the example but is not.

Sebastien-Ahkrin:
Is this example : https://931-integration-of-tanstack.react-science.pages.dev/stories/?path=/story/forms-form-tanstack--full-example 

looks good ?

stories/components/ts_form.stories.tsx
stropitek: > It's a bit confusing to pass this id property because it seems important in the example but is not.

@Sebastien-Ahkrin you can resolve this if this is fixed.

stropitek:
> Is this example : https://931-integration-of-tanstack.react-science.pages.dev/stories/?path=/story/forms-form-tanstack--full-example
> 
> looks good ?

Can you give some context?

stropitek:
> You wanted an example of form. I took some from peak by peak but I don't know if you want a bigger or no. Because the form from pbp is really big

We talked about getting an example from NMRium, not peak-by-peak

Sebastien-Ahkrin:
Après la réunion avec Daniel voici ce qu'il s'est dit : 

- on refait général settings de cheminfo/nmrium sur une autre pr.
- retire l'inline
- on retire le retour à la ligne gris et on met une plus grande séparation entre les sections (general settings)
- BUG: Select, si on selectione une valeur le form fail
- on désactive le fait que le bouton submit se désactive
- on peux voir plus tard pour focus la première erreur quand il y en a une
- Ajout du switch
- séparer les props : formGroupProps / inputProps
- retirer ts/tanstack. rename form en input, et ts_form en form

NB: Fieldset

src/components/form/components/input.tsx
```diff
@@ -0,0 +1,37 @@
+import type { ChangeEvent } from 'react';
+
+import { Input as FCInput } from '../components/form_components/input.js';
```
stropitek, line 3: Give it a full name like `FormGroupInput`

src/components/form/components/form_components/input.tsx
```diff
@@ -0,0 +1,68 @@
...
+import type { ChangeEvent } from 'react';
+
+import { useInputId } from '../hooks/use_input_id.js';
+
+interface InputProps {
+  error?: string;
+  required?: boolean;
+  id?: string;
+
+  inputGroupProps: {
```
stropitek, line 11: I would only put the formGroupProps in a separate property, not the actual input's props.

src/components/form/components/reset_button.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import { Button } from '../../button/index.js';
+import { useFormContext } from '../context/use_ts_form.js';
+
+interface ResetButtonProps {
+  children: ReactNode;
+}
+
+const StyledResetButton = styled(Button)`
+  background-color: #dd3f3f !important;
```
stropitek, line 12: You should not use hardcoded values.

src/components/form/components/form_components/select.tsx
```diff
@@ -0,0 +1,193 @@
...
+  renderButton?: (state: SelectPropsRenderButtonState<OptionType>) => ReactNode;
+  disabled?: boolean;
+  selected?: ID;
+  formGroupProps: Pick<
+    ComponentProps<typeof FormGroup>,
+    'helperText' | 'intent' | 'label' | 'className' | 'inline'
+  > & {
+    required?: boolean;
+  };
+  selectProps: Pick<BPSelectProps<OptionType>, 'filterable' | 'items'> & {
```
stropitek, line 105: Ditto, I would put those at the first level, not in a separate property.

src/components/form/components/submit_button.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { Button } from '../../button/index.js';
+import { useFormContext } from '../context/use_ts_form.js';
+
+type SubmitButtonProps = ButtonProps;
+
+export function SubmitButton(props: SubmitButtonProps) {
+  const form = useFormContext();
+  const [isSubmitting] = useStore(form.store, (state) => [state.isSubmitting]);
+
+  return <Button {...props} type="submit" disabled={isSubmitting} />;
```
stropitek, line 13: Maybe the submit button should have a primary intent by default.

src/components/form/utils/Label.tsx
```diff
@@ -0,0 +1,29 @@
...
+  children: ReactNode;
+  error?: string;
+  required?: boolean;
+  label?: string;
+  intent?: Intent;
+  labelFor?: string;
+  onClick?: () => void;
+}
+
+export function Label(props: LabelProps) {
```
stropitek, line 15: Let's discuss together about the purpose of this component.

src/components/form/utils/use_intent.ts
```diff
@@ -0,0 +1,5 @@
+import type { Intent } from '@blueprintjs/core';
+
+export function useIntent(error?: string): Intent | undefined {
```
stropitek, line 3: It's misleading to name this like a hook because it is not one.

stories/components/tanstack/full_example.stories.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+  & + & {
+    border-top: none;
+    margin-top: calc(-8px + -10px);
+  }
+
+  &:has(+ &) {
+    padding-bottom: 20px;
+  }
+`;
```
stropitek, line 33: Can you use very basic layout for the stories?

stories/components/tanstack/full_example.stories.tsx
```diff
@@ -0,0 +1,291 @@
...
+      }
+    }
+  }
+
+  /* tweaks for inputs in cell grid */
+  & .${Classes.FORM_CONTENT} .${Classes.CONTROL_GROUP} .${Classes.INPUT_GROUP} {
+    flex-shrink: 1;
+    flex-grow: 1;
+  }
+`;
```
stropitek, line 76: Ditto, this is too complicated and does not even look good

CleanShot 2025-08-29 at 15 03 41@2x

stories/components/tanstack/full_example.stories.tsx
stropitek, line 1: This story is complicated.

Try to narrow it down and focus on what you want to show with it.
Don't focus on layout, just on form functionalities.

stories/components/tanstack/ts_form.stories.tsx
stropitek, line 1: rename to remove the references to tanstack

src/components/form/components/numeric_input.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const id = useFieldId(field.name);
+  const error = useErrors(field);
+  const intent = useIntent(error);
+
+  function onChange(value: number) {
+    return field.handleChange(value);
+  }
+
+  return (
+    <Label
```
stropitek, line 28: Not sure why you decided to create a separate component for this case but not others.

I would not create simple `FormGroup` wrappers which mostly forward props.

Sebastien-Ahkrin:
First, I'm doing parsing in both stories with : `formSchema.parse(value)`

```tsx
const form = useForm({
    defaultValues,
    onSubmit: ({ value }) => {
      action('onSubmit')(formSchema.parse(value));
    },
    validationLogic: revalidateLogic({
      modeAfterSubmission: 'blur',
    }),
    validators: {
      onDynamic: formSchema,
    },
  });
```

Switch are in "Proof Of Concept" story : 

https://931-integration-of-tanstack.react-science.pages.dev/stories/?path=/story/forms-form-tanstack--proof-of-concept
under the "developer" input. 

I think i will remove the "full example" story because Proof Of Concept are currently doing that

src/components/form/utils/Label.tsx
Sebastien-Ahkrin: This component is non-usefull i removed it

src/components/form/utils/use_intent.ts
Sebastien-Ahkrin: i renamed it to "getIntent"

Sebastien-Ahkrin:
On /?path=/story/forms-form-inputs--numeric-input : If i remove everything, and send the form. The value is "0". What the best thing for you to happens here ?

src/components/form/components/input/checkbox.tsx
```diff
@@ -16,6 +16,8 @@ export function Checkbox(props: CheckboxProps) {
   function onChange(event: React.ChangeEvent<HTMLInputElement>) {
     const checked = event.target.checked;
     field.handleChange(checked);
+
+    return field.handleBlur();
```
stropitek, line 20: Pretending that the field blurred doesn't seem like a great way to handle the issue.

Either checkbox fields should be configured to revalidate on change, or we can change the global revalidation strategy to be on change for all types of inputs after the initial submit. I'd be in favor of the latter (change the global strategy).

src/components/form/components/input/numeric_input.tsx
```diff
@@ -14,17 +14,17 @@ interface NumericInputProps
...
+  const field = useFieldContext<string | undefined | null>();
   const id = useFieldId(field.name);
   const error = useErrors(field);
   const intent = getIntent(error);
 
-  function onChange(value: number, valueAsString: string) {
+  function onChange(_: number, valueAsString: string) {
     if (valueAsString === '') {
-      return field.handleChange(undefined);
+      return field.handleChange(null);
```
stropitek, line 24: Is passing `undefined` what reverts the value to its default value?

src/components/form/components/input/select.tsx
```diff
@@ -24,7 +24,9 @@ export function Select(props: SelectProps) {
 
   function onItemSelect(selected: SelectId | undefined) {
     if (!selected) return;
-    return field.handleChange(selected);
+    field.handleChange(selected);
+
+    return field.handleBlur();
```
stropitek, line 29: ditto

src/components/form/components/input/switch.tsx
```diff
@@ -19,6 +19,8 @@ export function Switch(props: SwitchProps) {
 
   function onChange(event: ChangeEvent<HTMLInputElement>) {
     field.handleChange(event.target.checked);
+
+    return field.handleBlur();
```
stropitek, line 23: ditto

src/components/form/components/input/checkbox.tsx
stropitek: [I made a major edit to my comment above]

src/components/form/components/input/numeric_input.tsx
Sebastien-Ahkrin: Yes, then if you want nothing in the input like you wanted before, we have to set "null"

Sebastien-Ahkrin:
We have a problem on the "City" input inside the example stories.
Before we set the default values to undefined but tanstack form say : https://tanstack.com/form/latest/docs/framework/react/guides/debugging#changing-an-uncontrolled-input-to-be-controlled

So i pass, "" as default values to correct the error in console. But, now we have "" in submitted values.
Do you think a way how to handle this correctly ? or maybe, "" is "ok" ?

src/components/form/components/input/numeric_input.tsx
stropitek: Why don't you pass an empty string then? The validation can transform an empty string to `undefined` then. The implementation of the NumericInput form component should stay close to a normal numeric input and not expect the `null` value.

PR #934, opened 2025-09-16
Docs general settings

Sebastien-Ahkrin:


stropitek (commented):
- The height taken by individual form elements is not consistent (input vs switch)
- Add example with checkbox
- The old stories now look broken. It should be possible to still use a regular blueprint form group layout with our form components. I suggest that you add a custom `Form` component which let's you choose an alternative layout without having to pass a prop to each form element.
- You can use container queries to decide of the layout of form elements in the "box" of each section. You can keep the media query as a fallback for browsers not supporting it.
- The label and the input don't look very centered, especially in individual input stories where it is completely off and does not look good.

lpatiny (commented):
I think the layout is now compact enough.

I spoke with you that if there are 2 lines of description the layout is broken but I don't think the solution is to keep only one line and not fix the issue.

Could you check what happens when the description, help, etc are long. I just edited the html and for the little help I think the text should be more compact in height.
For check boxes many lines also yield to some issues.

image

Sebastien-Ahkrin:
CleanShot 2025-09-29 at 14 14 41

Sebastien-Ahkrin:
> > I wonder if help text on checkbox should be left-align with the label ?
> 
> I think it looks good aligned with the checkbox

Yeah i think its better to be aligned with the checkbox and not with the text! For me, the current version looks good!

package.json
```diff
@@ -64,6 +64,7 @@
     "@tanstack/react-form": "^1.19.2",
     "@tanstack/react-table": "^8.21.3",
     "@tanstack/react-virtual": "^3.13.12",
+    "clsx": "^2.1.1",
```
targos, line 67: Can we avoid adding a dependency for just a simple use? In that case you can directly concatenate the classNames.

Sebastien-Ahkrin:
> Little comment about tanstack form (I know it's a bit out of context of this PR).
> 
> The way we create our form with it lead to hard navigation in code. There is no easy way from a form to navigate to `NumericInput` component for example.
> 
> ```ts
>         <form.AppField name="general.opacityDimmedSpectra">
>           {(field) => (
>             <field.NumericInput
>               fill
>               label="Opacity of dimmed spectra"
>               helpText="Value should be between 0 and 1."
>               step={0.1}
>             />
>           )}
>         </form.AppField>
> ```
> 
> CMD+Click on `form` -> CMD+Click on `useForm` -> search `NumericInput` into `createFormHook` config -> CMD+Click on it.
> 
> I understand it's the way tanstack designed their API but to mitigate the navigation problem, I suggest to create wrappers for our `formComponents` and `fieldComponents`, colocated with components binded with `createFormHook` config.
> 
> example for `numeric_input.ts`:
> 
> add
> 
> ```ts
> /**
>  * wrapper of NumericInput for tanstack boilerplate
>  * @see NumericInput
>   */
> export function NumericInputTS<F>(props: NumericInputProps & {form: F, name: string}) {
>   const {form, name, ...fieldProps} = props;
> 
>   return (
>         <form.AppField name={name}>
>           {(field) => <field.NumericInput {...fieldProps} />}
>         </form.AppField>
>   )
> }
> ```
> 
> We could also pass form by context.
> 
> @stropitek @targos, if you agree on this, let's make an issue from this comment.

I'm not certain we should add more complexity on this only for that ... For me, its not a good reason to make news components for that

stropitek:
As long as props are discoverable I think we are good, even if navigating to props would have been a nice to have.

tpoisseau:
So no issue to create, let's keep this tanstack form API like that (no wrapper).

Sebastien-Ahkrin:
new link are :

https://docs-general-settings.react-science.pages.dev/stories/?path=/story/forms-form-tanstack-generalsettings--general-settings

stropitek (commented):
https://docs-general-settings.react-science.pages.dev/stories/?path=/story/forms-form-tanstack-inputs--input

Image

PR #953, opened 2025-10-21
feat(Form): add ColorPicker component

Sebastien-Ahkrin:


targos:
Oh I was expecting something like that: https://952-create-a-colorpicker-app.react-science.pages.dev/stories/?path=/story/components-color-pickers--color-picker-dropdown-story

Don't you find it too big as a full component in the form?

Sebastien-Ahkrin:
> Maybe it can be a prop of the field to render a dropdown or the inline picker

You're right, it's better with a dropdown menu. I've set it up for 100% of uses cases.

PR #962, opened 2026-01-22
feat: support collapsing overflow items into dropdown menu

hamed-musallam:
Add "overflow" prop with two options:
- "wrap": items wrap (default)
- "collapse": collapse the items into dropdown menu

### DIFF
--- a/stories/components/toolbar.stories.tsx
+++ b/stories/components/toolbar.stories.tsx
@@ -426,6 +426,14 @@ export function HorizontalMore() {
   const selectedItem = itemsBlueprintIcons.find((item) => item.id === selected);
   return (
     <SplitPane>
+      <div>
+        {selectedItem && (
+          <div style={{ padding: 5 }}>
+            <p>Hello, World!</p>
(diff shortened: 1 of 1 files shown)


src/components/toolbar/Toolbar.tsx
```diff
@@ -91,13 +108,19 @@ export function Toolbar(props: ToolbarProps) {
...
+    vertical,
+  });
 
   // Work around wrong width on vertical flex when wrapping
   // In Chrome: recently fixed (https://bugs.chromium.org/p/chromium/issues/detail?id=507397)
   // In Firefox: work-around needed (https://bugzilla.mozilla.org/show_bug.cgi?id=995020)
   // In Safari: work-around needed
   useLayoutEffect(() => {
-    if (!vertical) {
+    if (!vertical || overflow !== 'wrap') {
```
stropitek: Was always wrap before, so should this be && ?
```suggestion
    if (!vertical && overflow !== 'wrap') {
```

PR #1001, opened 2026-05-04
feat: add svg export utils

tpoisseau:
* exportSVG
* exportSVGAsPNG

### DIFF
--- a/src/components/utils/svg.ts
+++ b/src/components/utils/svg.ts
@@ -1,0 +1,120 @@
+import { assert } from './assert.ts';
+
+export interface ExportSVGOptions {
+  /**
+   * CSS Selector to remove from the SVG.
+   *
+   * @example '[data-no-export="true"]' And add `data-no-export="true"` attribute to the elements you don't want in the exported SVG.
+   * @default undefined (Will not remove any elements)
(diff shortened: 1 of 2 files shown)


src/components/utils/svg.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  if (addBackground) {
+    ctx.fillStyle = backgroundColor;
+    ctx.fillRect(0, 0, canvas.width, canvas.height);
+  }
+
+  ctx.drawImage(img, 0, 0);
+
+  const png = await canvas.convertToBlob({ type: 'image/png' });
+  return png;
```
Sebastien-Ahkrin, line 119: is this really necessary to return the newly variable instead of returning `canvas.convertToBlob({ type: 'image/png' });` ?

Sebastien-Ahkrin (approved):

src/components/utils/svg.ts
tpoisseau: * It name things
* the await is done in the function (but could return-await)
  * If it throws, it is better to await in the right place to have a better trace.
  
It is not necessary, but I find it useful.

PR #1009, opened 2026-06-25
feat: update dependencies

targos:


package.json
```diff
@@ -23,7 +23,7 @@
     "createData": "npm run prepack && node scripts/createData",
     "dev": "echo Use npm run dev-app or npm run storybook && exit 1",
     "dev-app": "vite --open /pages/demo.html",
-    "eslint": "eslint . --cache --max-warnings=39",
+    "eslint": "eslint . --cache --max-warnings=45",
```
targos, line 26: I anticipate a comment here: the increase in warnings is due to changes in the React rules (mostly around useEffect).
IMO it's OK to increase the expected warnings if they are not caused by changes in the source code and are not trivial to fix.

src/components/table/use_table_scroll.ts
```diff
@@ -63,7 +63,7 @@ export function useTableScroll<TData extends RowData>(options: {
           options?: TableVirtualScrollIntoViewOptions,
         ) {
           const element = scrollRef.current?.querySelector(
-            `tr[data-row-id="${id}"]`,
+            `tr[data-row-id="${CSS.escape(id)}"]`,
```
tpoisseau, line 66: I did not know this helper. A variant adapted to template string would be great.

```js
selector`tr[data-row-id="${id}"]`
```

I'll create an issue so we take time to add this helper to react-science if you agree.

tpoisseau (approved):
LGTM.

I had issue to load cloudflare page (disabling cache was enough). local build and preview seems OK.

src/components/table/use_table_scroll.ts
targos: I didn't know either. The suggestion came from a new "unicorn" rule.
OK for a helper in principle, but then the rule will suggest the wrong thing.

tpoisseau: > but then the rule will suggest the wrong thing.

You may right. I'll create the issue with this reserve.

tpoisseau: https://github.com/zakodium-oss/react-science/issues/1010

PR #1013, opened 2026-06-26
feat: provide `AppForm` component

tpoisseau:
for unified and simplified form declaration.

It deprecate `Form` (with migration instructions).

Closes: https://github.com/zakodium-oss/react-science/issues/1012

docs: add stories for AppForm

* show regular use case
* show use-case with meta submit

### DIFF
--- a/stories/components/form/tanstack/field-groups.stories.tsx
+++ b/stories/components/form/tanstack/field-groups.stories.tsx
@@ -4,19 +4,19 @@ import type { ReactNode } from 'react';
 import { action } from 'storybook/actions';
 import { z } from 'zod';
 
-import type { FormProps } from '../../../../src/components/form/components/input_groups/form.js';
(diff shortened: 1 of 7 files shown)


stories/components/form/tanstack/field-groups.stories.tsx
```diff
@@ -58,24 +58,15 @@ export const SVGTextStyle: StoryObj<
...
-          <FieldGroupSVGTextStyleFields
-            form={form}
-            fields="textStyle"
-            label={label}
-            previewText={previewText}
-          />
-          <form.SubmitButton>Save</form.SubmitButton>
-        </Form>
-      </form.AppForm>
+      <AppForm form={form} layout={layout} style={{ margin: '5px' }}>
```
Sebastien-Ahkrin: Can you use Styled to create the style like the other stories ? 
I think its better to use it everywhere

Sebastien-Ahkrin (approved):

PR #1022, opened 2026-07-03
feat(AppForm): add `onSubmitStopPropagation` flag

tpoisseau:
Closes: https://github.com/zakodium-oss/react-science/issues/1018


Sebastien-Ahkrin (commented):
I think, we should at least have one stories that show this props.
If not, in a couple of month or year we will forgot this

Sebastien-Ahkrin (approved):

PR #1028, opened 2026-07-20
fix: deprecate assert methods

tpoisseau:
in favor of `@zakodium/utils`

* `assertUnreachable`
* `assertNotNull`
* `assert`

### DIFF
--- a/src/components/color-picker/preview/color_config.ts
+++ b/src/components/color-picker/preview/color_config.ts
@@ -1,0 +1,21 @@
+import type { GradientScaleName } from '../gradient_select/index.ts';
+
+export type ColorConfig =
+  | { kind: 'fixed'; color: string }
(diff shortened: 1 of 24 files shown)


src/components/color-picker/preview/color_config.ts
```diff
@@ -0,0 +1,21 @@
+import type { GradientScaleName } from '../gradient_select/index.ts';
+
+export type ColorConfig =
+  | { kind: 'fixed'; color: string }
+  | { kind: 'fixedGradient'; gradient: GradientScaleName };
+
+export function assertColorFixed(
+  color: ColorConfig,
+): asserts color is { kind: 'fixed'; color: string } {
```
targos: Nit: give a name to the type to avoid repeating it. Same below.

targos (approved):

PR #1030, opened 2026-07-22
fix: adjust icon margin when render aside a tag

tpoisseau:
before it was always adjusted.
But there is weird edge case the `margin-right: 0px;` is ignored because blueprint uses higher specificity. It is un-noticable on stories.

In some app, we use css `@layer` to import react-science and blueprint css stylesheets. In this case `margin-right: 0px;` always apply and breaks all icons alignment (because no-layer wins over layer).

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -42,25 +42,27 @@ const ButtonTag = styled(Tag)`
 
 interface InnerButtonProps {
   isIconButton: boolean;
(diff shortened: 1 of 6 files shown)


stropitek (commented):
Won't the issue remain for buttons with a tag when we use layers?

Should we use layers in the stories to ensure the bug is fixed?

stropitek (approved):

PR #1039, opened 2026-08-03
refactor: remove deprecated assertion helpers

targos:
BEGIN_COMMIT_OVERRIDE
refactor: remove deprecated assertion helpers

BREAKING CHANGE: Removed deprecated assertion helpers. Use `@zakodium/utils` instead.
END_COMMIT_OVERRIDE


src/components/utils/index.ts
```diff
@@ -1,4 +1,3 @@
-export * from './assert.js';
 export * from './debounce.js';
 export * from './throttle.js';
 export * from './svg.js';
```
Sebastien-Ahkrin, line 3: Maybe we should create those on utils too ?

Sebastien-Ahkrin (approved):

src/components/utils/index.ts
targos: Maybe, but only if we need them somewhere else.

PR #1042, opened 2026-08-12
feat(validation): add `coerceNumberInput` on validation utils

Sebastien-Ahkrin:
Refs: https://github.com/cheminfo/nmrium/issues/4308


.npmignore
Sebastien-Ahkrin, line 1: Because of test inside the src folder (I wanted to keep tests file along the source file) I had to exclude `**/*.test.*` directly in `.npmignore` for the CI.

Tell me if its ok for you

package.json
```diff
@@ -10,7 +10,8 @@
   "files": [
     "lib",
     "src/components",
-    "styles/preflight.css"
+    "styles/preflight.css",
+    "!src/components/**/*.test.*"
```
targos:
```suggestion
    "styles/preflight.css"
```

You don't need this with the change in .npmignore

.npmignore
targos: This file should be in `./src` and based on https://github.com/cheminfo/generator-cheminfo/blob/main/generators/typescript/templates/npmignore

src/components/form/utils/validation/coerce_number_input.ts
targos, line 1: snake_case

targos, line 1: Also you need to export this file from `index.ts`

targos (approved):

PR #1048, opened 2026-08-14
fix: change transform to codec in coerceNumberInput

Sebastien-Ahkrin:
To be working in nmrium, we have to provide a way to decode and encode the number / string


targos (approved):
Try to add a test?
This needs to be merged as `fix` or you won't have a release

targos:
I merge, but please add a test in a new PR.

